PR #7, opened 2023-06-02
Add Comment.insertReply method

samalone:
This is a quick recursive method to insert a reply into a Comment. You may want to adapt this to your own code style.

### DIFF
--- a/Mlem/Models/User-Interactable/Comment.swift
+++ b/Mlem/Models/User-Interactable/Comment.swift
@@ -51,4 +51,20 @@ struct Comment: Codable, Identifiable, Hashable
 
     let childCount: Int?
     var children: [Comment]
+    
+    func insertReply(_ reply: Comment) -> Comment {
+        if id == reply.parentID {
+            var result = self
+            result.children.append(reply)
+            return result
+        }
+        else if children.isEmpty {
+            return self
+        }
+        else {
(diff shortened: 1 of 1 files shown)


samalone:
One thing I worry about is that Comment also has a `childCount` field. I'm not sure how this field is used or why it's an `Int?` instead of an `Int`, but if it's supposed to be the same as `children.count`, then my code won't work correctly since it's not updating `childCount`.

Could `childCount` be redefined as:

```swift
var childCount: Int {
    children.count
}
```

That would ensure that it's always accurate.

samalone:
An Xcode search shows that `Comment.childCount` is never used, so perhaps it should be deleted.

buresdv:
`childCount` is a vestige from my old way of parsing the JSON response, where I had to define every single property from the JSON response. Since then, I have learned how to ignore properties, so that property is now relly unused.

Good eye! I will remove it 😉

buresdv:
Oh, would you be up for one more challenge? If you're not sick of it, of course 😂

samalone:
Sure, I'm game for more! I'm teaching several sailing classes this weekend, so I might not be as quick on the next one.

buresdv:
Great! Check out #12, and enjoy the sailing 😊

Issue #12, opened 2023-06-03
Implement function for replacing comments in the comment tree 


buresdv:
## Introduction
As it is now, in order to rate a comment, the entire view must be force-reloaded to apply changes to the comment items

## Requirements
There should be a function which will accept an updated `Comment` struct. it will then traverse the comment tree, find the old `Comment`, and replace it with the new comment, preserving the current contents of the `children` property.

## Discussion
Much like with the previous system of creating replies, in order for a user's vote on a comment to show up, after each voting operation, Mlem needs to re-pull the entire comment tree from the API, parse it, and force-reload the view to get the vote to show up. This is very inconvenient, as it can take a couple of seconds to vote on a post; and if the user presses any of the voting buttons quickly in succession without waiting fo the updated comment tree to become available, Mlem will crash.

This improved system would instead replace only the relevant comment in the tracker, which would eliminate the time and energy-demanding system of pulling the entire tree again. It would be much more seamless and not prone to crashes.

This system would later be used for other types of comment manipulation, such as saving comments, editing their contents and deleting them. This is why I believe it would be more convenient to replace the entire `Comment` node in the tree, instead of just manipulating its various properties.

It might be more trivial compared to #5.

@samalone

PR #16, opened 2023-06-03
Add Comment.replaceReply and .removeReply

samalone:
I added both Comment.replaceReply() and Comment.removeReply().
removeReply might be handy if someone decides to delete their comment.
Closes https://github.com/buresdv/Mlem/issues/12

### DIFF
--- a/Mlem/Models/User-Interactable/Comment.swift
+++ b/Mlem/Models/User-Interactable/Comment.swift
@@ -67,4 +67,46 @@ struct Comment: Codable, Identifiable, Hashable
             return result
         }
     }
+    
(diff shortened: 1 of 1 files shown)


buresdv:
Thank you, does this also work for top-level comments, i.e. those comments with no parentID?

samalone:
Yes, it should work for top-level comments.
Be aware that the replacement can also change the reply's children!
This could be handy, but it could also be surprising.
If you don't want the children to change, be sure the replacement has the same children as the original.

buresdv:
Alright, thanks for the info! This is exactly what I needed 😊

buresdv (approved):
Looks good

PR #20, opened 2023-06-04
Fix - external link handling

mormaer:
This PR addresses the issue described [here](https://github.com/users/buresdv/projects/4/views/1?pane=issue&itemId=29858951).

I have updated things so that we stay in a `SwiftUI` context via a representable wrapper rather than interacting directly with `UIKit`.

Requests to open a URL by the application are now caught at the `ContentView` level, in the future we can pass any URL we want handled into the environment and the behaviour of modally presenting the `SafariView` should occur.

I've added a TODO as there should be some consideration made of how URLs which we can determine are linking to communities/posts etc should be handled, as a better UX would be for the user to be navigated within the application to our native views as opposed to treating all links as external.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		503BA26D2A2C93C70052516C /* SafariView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 503BA26C2A2C93C70052516C /* SafariView.swift */; };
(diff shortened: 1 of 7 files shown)


buresdv:
Issue was fixed by replacing `NavigationLink` with a `NavigationStack`

buresdv:
I'd be more than happy to implement any other changes from this PR, could you just open a new one?

mormaer:
> I'd be more than happy to implement any other changes from this PR, could you just open a new one?

I can do - I'll pull in the latest changes from `master` when I have time tomorrow and rebase to see if there are any conflicts - nice to see the new navigation API was able to resolve it though 🎉

Issue #13, opened 2023-06-03
* App crashes when loading sticky


ExperiBass:
**Describe the bug**
Attempted to visit the [sticky on lemmy.ml.](https://lemmy.ml/post/971001)

**Expected Behavior**
The post to load correctly.

**Actual Behavior**
The app crashed.

**To Reproduce**
Tap on the post.

**Mlem Information:**
version is 0.0.6, i didnt see a build number? didnt happen on 0.0.5

buresdv:
Man, that post is causing nothing but problems. I'll look into it again, it seems to be a problem with the data fetching function

ExperiBass:
good luck, id help but xcode doesnt seem to like me :C

mormaer:
The crash here is caused by the size of the response exceeding the maximum our task allows, I had a look and we don't currently define a `.maximumMessageSize`, so the default value is being used: 1048576 bytes.

That post currently has 408 comments and counting so it's response is huge compared to most others - the crash is resolved by increasing that value for example:
```swift
task.maximumMessageSize = 1048576 * 2
```
This isn't a robust solution as obviously the application would crash if it encountered a post with a larger response than that now 🙃 

I see there's a ticket for [migrating to the REST api](https://github.com/users/buresdv/projects/4/views/1?pane=issue&itemId=29851430) which I'd imagine will make this a non-issue in the future.

If moving to REST instead of websockets is something that will happen soon, it could be worth the _quick and dirty_ fix of increasing the maximum as above to resolve the crash, rather than sink time into identifying a robust solution that will be redundant in the near future.

buresdv:
Thank you for the insight! Yes, I'm planning to migrate to the REST API ASAP, but the process has been a bit rocky.

I have a fire under my ass because the WebSocket API will be removed sometime this month, so I'll have to do the migration until then.

I think it will be safe to increase the max task size temporarily. Originally, I wanted to have a system where it would load only a few top-level comments with the rest on-demand, but I have other priorities at the moment.

mormaer:
as noted above I've opened a PR with that small change ^

buresdv:
Thank you, merged :)

PR #23, opened 2023-06-04
fix - increase maximum message size

mormaer:
As discussed [here](https://github.com/buresdv/Mlem/issues/13) the application crashes when attempting to load to many comments.

As the migration to REST is in progress this temporary fix will mitigate the crash win the short term until the web socket implementation is replaced.


buresdv (approved):
Thank you very much, looks good!

PR #46, opened 2023-06-06
present URLs from SwiftUI context

mormaer:
This PR contains the changes originally proposed in #20.

In addition to those there is one additional change afa3bf6a20c22a4325a070a15e8232a6710f7f16, which prefers using `.fullScreenCover` for presentation of the Safari representable view. After looking at it again use of `.sheet` was not ideal as it caused the navigation bar to behave strangely on collapse.

With this change presentation of URLs its handled at the root of the application which should allow any child views to display a URL by passing the acton into the environment as seen in 59b045d41f8c4e184e93b954827242a8acf8bc9d.

I've added a TODO for considering how deep linking could be implemented - as remaining in the application when tapping links to posts/comments etc should be preferred.

---

The only visible change in behaviour is that links are now modally presented/dismissed instead of pushed/popped


buresdv:
I've been thinking about this for a few days and I still think the best option to display the webpage is to have the SafariViewController being pushed on top of the NavigationStack using the UIKit `present` method. Now that we fixed the issue with the navigation being messed up after closing the preview, I see no reason to transition away from it. 

I know it is not the most "SwiftUI" way, but it's the most native way. Other apps written in SwiftUI, namely IceCubes, try to emulate this behavior using SwiftUI, but it doesn't feel the same.

I think you did great work here, I like the contribution and think it is a great basis for foture link handling, but could we keep the UIKit `present` way of showing the webpage?

mormaer:
@buresdv - thanks for the response

If we leave this PR open I'll have another look at it, I'd say that calling out to `UIKit` in this way is OK in our current position, but we may need to revisit that approach in the future.

One of the benefits of remaining inside the `SwiftUI` context is that it ~will~ _should_ automatically adapt for the platform it runs on, if the application was to run on iPadOS/macOS at some stage later use of the `keyWindow` may behave strangely, but that is a bridge that can be crossed when we come to it 👍 

---

If I'm understanding your above request correctly, you'd like to keep the approach of handling the links via the environment (so that going forward we can pass links we want handled into it from anywhere), but instead of `.fullScreenSheet`, we'd revert to using the original `.present` approach at that last stage?

If my understanding is correct, I'll make that change and push a new commit onto this branch 👍

buresdv:
> One of the benefits of remaining inside the SwiftUI context is that it will should automatically adapt for the platform it runs on, if the application was to run on iPadOS/macOS at some stage later use of the keyWindow may behave strangely, but that is a bridge that can be crossed when we come to it 👍

That's a great point, since multiplatform support is in the pipeline. But I don't think it's time-sensitive ATM, since it's not a priority, so, as you said, we can cross that bridge when we come to it :)

> If I'm understanding your above request correctly, you'd like to keep the approach of handling the links via the environment (so that going forward we can pass links we want handled into it from anywhere), but instead of .fullScreenSheet, we'd revert to using the original .present approach at that last stage?

You understood it perfectly.

nick-lawson:
Nice, I'll get that done soon!

mormaer:
Nice, I'll get that done soon 👍 

(if you got a notification for another commenter it's me, turns out I'm logged in on my main account in mobile 😅)

mormaer:
@buresdv - I've updated this PR with some new commits, we're now back to using the original presentation via the `.keyWindow` approach 👍 

In addition to the I moved the link handling (although there's not really any _handling_ yet) into it's own class, thought being that it gives us a single place to add future logic to handle deep linking for community links etc...

The PR is rebased onto `master` and I updated the links in `SettingsView` to use the environment approach to save the references to `UIApplication` in there

buresdv (approved):
Great job, thank you very much! 👍

PR #47, opened 2023-06-07
Add bio and username prepended with @ symbol to User View

thx1766:
I modified the user view to display the username prepended with "@". I also added the user bio to the view.

Note: on my local I ran into some weird errors with modifying an isolated variable in the main actor. I worked around it with a hack, but without reaching out to the development team about my environment I figured I'd just submit the relevant changes and leave out my weird hack.


buresdv:
Thank you for the contribution! There was a complete overhaul of the user view just now, so unfortunately, I can't mere this PR because the changes have already been implemented.

Issue #42, opened 2023-06-05
iPad support


buresdv:

thx1766:
Started here https://github.com/buresdv/Mlem/pull/48

thx1766:
Before adding iPad to supported destinations:
![image](https://github.com/buresdv/Mlem/assets/6318596/ecbf469b-4bff-4846-8be8-be0f0bd46dff)
![image](https://github.com/buresdv/Mlem/assets/6318596/b684a785-4895-40c6-959f-69d5a01bd392)
![image](https://github.com/buresdv/Mlem/assets/6318596/36ec0d8e-78bb-4bba-b4b1-332ef0aea000)

![image](https://github.com/buresdv/Mlem/assets/6318596/f0e7e967-5562-43b2-9536-2862aaa72343)
![image](https://github.com/buresdv/Mlem/assets/6318596/04b21daa-a8bb-4781-b513-ea1617a578a7)
![image](https://github.com/buresdv/Mlem/assets/6318596/3e4c710a-1368-4ca8-96ac-cc42b838dc3f)
![image](https://github.com/buresdv/Mlem/assets/6318596/2fea411d-4808-467a-9b6a-94a513488053)

After:
![image](https://github.com/buresdv/Mlem/assets/6318596/c73ad69e-2d48-4cac-81e4-15067af7d6f5)
![image](https://github.com/buresdv/Mlem/assets/6318596/8627600f-26a0-4a1d-a161-9f92d41cd952)
![image](https://github.com/buresdv/Mlem/assets/6318596/d2a82249-288b-4674-9e05-c447a44589b9)

hugpeter:
if you are frustrated with mlem not supporting ipad, memmy is an alternative iOS app that currently does support it. memmy is not as far along in development as mlem and I definitely prefer mlem at this point in time, not sure what the hold up is for mlem on this if someone could educate me.

PR #48, opened 2023-06-07
Add iPad to Supported Destinations

thx1766:
Add iPad to Supported Destinations


andrew-vanderbilt:
Making it a giant iPhone app would be detrimental to the experience of the user. Unless the app undergoes some redesigns to utilize the larger screen real-estate, I fail to see how just enabling the build iPad family target is a benefit.

isametry:
…More detrimental than an upscaled 750px iPhone app surrounded with black bars? This has to be trolling right?

Even if precisely ZERO further enhancements are added to take advantage of an iPad screen, this change singlehandedly takes the app from “unbearable on iPad” to “useable on iPad”.

ShadowJonathan:
Multiplatform support is tracked under #40, this is a PR from an external contributor, please don't act or banter like this is an officially-approved PR yet, as we have yet to review it.

There's been discussion on the best way to do iPad support the last few days, and @thx1766 was a bit eager to send a PR ahead of time.

isametry:
I'm sorry, you're right, this is not the place for discussion. I've stripped the unnecessarily agitated parts from my previous comment.

thx1766:
Here are some screenshots of the current state of Mlem on the iPad, assuming we took the current main branch and deployed it to an iPad.

The first two screenshots are portrait and landscape views of the simulator showing the iPhone-shaped result.

The other images are from my fiancé's physical iPad running multiple apps with Stage Manager. Here Mlem is resizable, but not necessarily ideal.

Simulator Screenshot - iPad (10th generation) - 2023-06-10 at 11 40 44
Simulator Screenshot - iPad (10th generation) - 2023-06-10 at 11 40 21
IMG_9ABB4EDF7EC5-1
IMG_7EBAFDF01532-1
4c5)
IMG_C92E699F2119-1

This last screenshots show the settings tab with nested navigation, which swiftUI renders in a sensible way on iPhone but seems silly and requires extra taps to navigate around on an iPad. If there are a lot of settings that can't be managed on one iPad sized page, it might make sense to group them this way, but it seems like a waste of screen real estate to me. Just my opinion for the design team to take into consideration.
IMG_E6EC40C0DE40-1

IMG_C2CEA96CAFA9-1

buresdv:
Thank you, what if you apply the changes in this PR to the current state of the master branch? How does that look?

thx1766:
Adding the iPad supported destination, we get a much more elegant settings menu.
IMG_AD017CA68A5E-1
IMG_AF0FE96DE023-1

thx1766:
Here's the simulator showing landscape settings page in full screen mode.
Simulator Screenshot - iPad (10th generation) - 2023-06-10 at 11 58 36

thx1766:
Closing this out as this is not a priority currently.

buresdv:
Thank you very much for the pictures, I think that even if we don’t change much, the experience is much better.

Can you please post all these pictures to #42? And we should continue the discussion there.

PR #51, opened 2023-06-07
fix #50 - avoid force unwrapping urls

### LINKED ISSUE #50: Fix the crashing
https://drive.google.com/file/d/1tRf8_yrl8hVReBb0SPm1DRKEFQy8_l-N/view?usp=share_link

Mostly has to do with this function: https://github.com/buresdv/Mlem/blob/master/Mlem/Logic/Networking/Infinite%20Loader.swift

mormaer:
This PR addresses #50 

As noted in the above issue the crash reports highlight a runtime error:

`Unexpectedly found nil while unwrapping an Optional value`

The changes [here](https://github.com/buresdv/Mlem/compare/master...mormaer:fix-50-force_unwrap_url_crash?expand=1#diff-c58a875406d9325f1c6fc2c6497b428aa2a1fdb8d0ac6e48e5774e1bafd957acR18) remove the force unwrapping (`!`) on the URLs, if an invalid `baseInstanceAddresss` parameter is fed in we will now throw the error instead of crashing.

You can test this fix with the following steps:
- [ ] On the existing build enter an invalid address on the add account page, eg `beehaw   .org`, note the whitespace
- [ ] Enter credentials and tap log in
- [ ] Confirm the app crashes
- [ ] Build this branch and repeat the above steps
- [ ] Confirm you are shown the `unable to connect...` error message without crashing

---

Added two additional commits to address other crashes for force unwrapping in the logs.

Of the remaining issues that the logs identify:
- a few are low-level Swift errors which are difficult to resolve without steps to replicate
- one is in `convertResponseDateToDate` which appears to have been mitigated by previous changes, the reports appear to be from app version `0.0.2`
- one is linked to the current Safari view implementation which should be resolved by #46

### DIFF
--- a/Mlem/Logic/Networking/Get Correct URL to Endpoint.swift
+++ b/Mlem/Logic/Networking/Get Correct URL to Endpoint.swift
@@ -7,45 +7,34 @@
 
 import Foundation
 
-fileprivate enum EndpointDiscoveryError: Error
-{
+enum EndpointDiscoveryError: Error {
     case couldNotFindAnyCorrectEndpoints
 }
 
-func getCorrectURLtoEndpoint(baseInstanceAddress: String) async throws -> URL
-{
+func getCorrectURLtoEndpoint(baseInstanceAddress: String) async throws -> URL {
     var validAddress: URL?
     
-    let possibleInstanceAddresses: [URL] = [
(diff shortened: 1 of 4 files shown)


buresdv (approved):
Great contribution, thank you

PR #72, opened 2023-06-10
fix #69 - avoid multiple `loadInfiniteFeed` calls

### LINKED ISSUE #69: posts loop on any sort other than Hot
**Describe the bug**
What did you do to trigger the bug?
Went on a sort that’s not hot

**Expected Behavior**
What did you expect to happen?
For the posts to sort by date

**Actual Behavior**
What actually happened?
Some of them sorted by date, but most didn’t show up and instead the first few started looping

**To Reproduce**
What steps do I need to take to trigger the bug?
 - log in through beehaw (not sure if it happens on other instances)
 - Toggle on a sort other than hot in the homepage 

**Mlem Information:**
Version + Build number
Beta 0.0.6
If this started happening recently, write which version introduced the bug

**Additional Context**
Any other relevant information

mormaer:
This PR addresses the bug raised in #69 

While working on another branch (the API changes) I ran into this issue during testing, I have resolved it on that branch but thought it'd be worth a separate PR to resolve it in `master` rather than wait on that work wrapping up.

The issue was being caused by the `.onChange` behaviour in SwiftUI (shakes fist). When a user selected a new sorting option the callback to report the new value happens multiple times, as a result the existing logic would start loading the posts feed multiple times, which would result in duplicates being added to the `postTracker`.

The resolution is to use a manual `Binding` initialiser which only calls once on selection.

*Note: If you re-select the same feed type the behaviour now re-loads the feed, this could be avoided with a `guard` if we prefer, but this does mean if you're a few pages down your feed and re-select your option you're returned to page 1/top which may be desirable behaviour, if not let me know an I can add the `guard` to stop it*

You can confirm this fix with the following steps:
- [ ] Run the current `master` branch
- [ ] Navigate into the post feed
- [ ] Change to a sort option other than the default
- [ ] Begin scrolling down your feed
- [ ] Confirm you see the same posts repeat (if you are debugging in Xcode you'll see SwiftUI complaining about duplicate ids too)
- [ ] Build and run this branch
- [ ] Repeat the above steps
- [ ] Confirm the feed no longer repeats

### DIFF
--- a/Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
+++ b/Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
@@ -151,37 +151,6 @@ struct CommunityView: View
                                     }
                                 }
                             }
-                            .onChange(of: selectedSortingOption, perform: { newValue in
-                                Task
-                                {
-                                    print("Selected sorting option: \(newValue), \(newValue.rawValue)")
-
-                                    postTracker.posts = .init()
-                                    postTracker.page = 1
-
-                                    if community == nil
-                                    {
-                                        
-                                        if postTracker.posts.isEmpty
-                                        {
-                                            postTracker.isLoading = true
-                                        }
-                                        
-                                        await loadInfiniteFeed(postTracker: postTracker, appState: appState, community: nil, feedType: feedType, sortingType: selectedSortingOption, account: account)
-                                        postTracker.isLoading = false
-                                    }
-                                    else
-                                    {
-                                        if postTracker.posts.isEmpty
-                                        {
-                                            postTracker.isLoading = true
-                                        }
-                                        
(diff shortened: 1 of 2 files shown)


buresdv:
> Note: If you re-select the same feed type the behaviour now re-loads the feed, this could be avoided with a guard if we prefer, but this does mean if you're a few pages down your feed and re-select your option you're returned to page 1/top which may be desirable behaviour, if not let me know an I can add the guard to stop it

I think the best way to approach this would be to make the current selected sorting option disabled. What do you think?

mormaer:
> I think the best way to approach this would be to make the current selected sorting option disabled. What do you think?

I agree - good shout!

I've added a custom view `OptionButton` which provides that behaviour, it's generic on an `Equatable` type too, so we could re-use it elsewhere for other sets of options if we wanted that again 👍

buresdv (approved):
Amazing, thank you for another great contribution 😄

PR #76, opened 2023-06-10
Use version 3 first

lil5:
1. Version 1 it used first, this pr reverses it to try api v3 first.
2. If the app is simulated, http is added as a possible protocol if https doesn't work.


justincmoy:
Thanks for this! The `#if targetEnvironment(simulator)` block was especially helper when using a local HTTP server.

lil5:
> If I understand my code properly, they are all fired at once and then it waits for one of them to return a valid response code, so it shouldn't matter what order they're in.

Nope it tries one by one, if nothing works it errors out.

If your using v3 api calls already it might be good to:
1. First check using v3
2. If in a simulated environment check http v3 (thats what those `#if` statements are) 
3. Then if that doesn't work check v1 and show an error message saying the lemmy server is too old

@buresdv If that makes sense I can change this PR

buresdv (approved):
Looks good

PR #95, opened 2023-06-10
`APIClient` with modelled requests and responses

mormaer:
This PR adds an `APIClient` and models the requests and responses we use in the application.

As has been discussed in the development chat this is the first step, and I profusely apologise for the size of the PR I present to thee...

Due to the size of this PR I have split it into two parts:

273bc34282138a6fff14156c5652d4070175d69c - Adds the client, request protocol and the API models
6512451c33048b4e3534cc04a7e9ec8820b81556 - Refactors many sections of the application to use the `APIClient`

This gives us the opportunity to decide to bring the client in first (this would unblock @ShadowJonathan, and allow them to start modelling other areas of the API) and delay the refactoring changes if we feel they are too far reaching in scope.

**I would suggest that if we can verify the behaviour of the application and confirm there have been no regressions then bringing both parts in together makes sense as it saves drip feeding the other parts in over a week or so, which would continue to impact other work in progress, for what would be the same end result.**

With all that said, here's an overview of the approach this implements:

### `APIRequest` protocol

This protocol defines the conformance any request must have and consists of an endpoint, method, body data (`POST`) or query parameters (`GET`). In addition to this it has an `associatedtype` which represents the model this request expects to receive from the API.

Here is an example of defining a model:
```swift
struct LoginRequest: APIRequest {
    
    // 1. Associates the response model expected from the API
    typealias Response = LoginResponse
    
    let endpoint: URL
    let method: HTTPMethod
    
    // 2. Defines the body contents, these conform to `Encodable` to convert directly to data
    struct Body: Encodable {
        let username_or_email: String
        let password: String
    }
    
    init(instanceURL: URL, username: String, password: String) throws {
        // 3. On initialisation we build the endpoint...
        self.endpoint = instanceURL
            .appending(path: "user")
            .appending(path: "login")
        do {
            // 4. ... and then create our body using the values passed to us
            let data = try JSONEncoder().encode(
                Body(username_or_email: username, password: password)
            )
            
            // 5. set our method, in this case we'll post the encoded values from above
            self.method = .post(data)
        } catch {
            throw APIRequestError.encoding
        }
    }
}

// 6. Define the response, in the case it is a basic string representing our JWT
struct LoginResponse: Decodable {
    let jwt: String
}
```

<details>

<summary>Alternative example of a `GET` request</summary>

```swift
struct GetCommentsRequest: APIRequest {
    
    // 1. Associates the response model expected from the API
    typealias Response = GetCommentsResponse
    
    let endpoint: URL
    let method: HTTPMethod
    // 2. This time we define that we have items...
    let queryItems: [URLQueryItem]
    
    init(
        account: SavedAccount,
        postId: Int,
        maxDepth: Int = 15,
        type: FeedType = .all
    ) {
        // 3. Construct our query items with the values provided...
        let queryItems: [URLQueryItem] = [
            .init(name: "auth", value: account.accessToken),
            .init(name: "post_id", value: "\(postId)"),
            .init(name: "max_depth", value: "\(maxDepth)"),
            .init(name: "type_", value: type.rawValue)
        ]
        
        self.queryItems = queryItems
        // 4. Build our endpoint and include the query items...
        self.endpoint = account.instanceLink
            .appending(path: "comment")
            .appending(path: "list")
            .appending(queryItems: queryItems)
        // 5. Set our method, in this case `GET`
        self.method = .get
    }
}

// 6. Define the response model which will be an array of comments...
struct GetCommentsResponse: Decodable {
    let comments: [APICommentView]
}
```

</details>

### `APIClient`

At present the client is simple and holds no state, it consists of one primary method:

```swift
func perform<Request: APIRequest>(request: Request) async throws -> Request.Response
```

As this is generic it can be passed any conformer of `APIRequest`, it then does the job of converting it's values into a `URLRequest` making the call to the network, modelling the response and returning the value to the caller of the method.

### Usage

With the above two parts in place, performing network calls boils down to creating your request and passing it the client and then using the response:

```swift
let request = try LoginRequest(instanceURL: instanceURL, username: usernameOrEmail, password: password)
let response = try await APIClient().perform(request: request)
print(response.jwt) // prints the `String` returned from the API
```

---

The refactoring part of the PR swaps the existing networking to use this new approach, and reads the original values from the new modelled responses. In addition to this there are some places where the code was simplified/broken up for readability.


Mlem/API/APIClient.swift
```diff
@@ -0,0 +1,94 @@
...
+            
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+            if let date = formatter.date(from: string) {
+                return date
+            }
+            
+            // TODO: is there somewhere that covers all the formats lemmy instances may return?
+            // for now we'll return the current date until we know what other formats will
+            // be encountered.
+            return .now
```
ShadowJonathan, line 34: Imo, throw here, just for correctness' sake

Mlem/API/APIClient.swift
```diff
@@ -0,0 +1,94 @@
...
+        })
+        return decoder
+    }
+}
+
+enum HTTPMethod {
+    case get
```
ShadowJonathan, line 41: There's three methods; GET, POST, and PUT

POST and PUT work the same in that they just send a JSON body, but GET serialises everything to its query parameters

Maybe its better to generalise data into this field, and then remove the query items bit on the APIRequest, because it'll never get used for PUT/POST, and vice versa, json body data will never be used for GET

Mlem/API/APIRequest.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+
+enum APIRequestError: Error {
+    case encoding
+}
+
+protocol APIRequest {
+    associatedtype Response: Decodable
+    
+    var endpoint: URL { get }
```
ShadowJonathan, line 17: Wouldn't it be better to make these immutable? Not sure, but since we're basically never editing an APIRequest after creation, swift could use the hint to maybe optimise the struct/handling?

ShadowJonathan (commented):
Some notes while I do a "bed time reading"

Mlem/Logic/Time Parser.swift
```diff
@@ -7,29 +7,29 @@
...
-    
-    #warning("TODO: Remove this in favor of figuring out WTF is crashing it")
-    if convertedDate == nil
-    {
-        convertedDate = Date()
-    }
-    
-    return convertedDate!
-}
+//func convertResponseDateToDate(responseDate: String) -> Date
```
ShadowJonathan, line 10: Maybe remove entirely?

Mlem/Logic/Networking/Infinite Loader.swift
```diff
@@ -8,68 +8,43 @@
...
-        appState.alertMessage = "Your network conneciton is either not stable enough, or the Lemmy server you're connected to is overloaded.\nTry again later."
-        appState.isShowingAlert.toggle()
         
-        //throw ConnectionError.failedToSendRequest
+        await MainActor.run {
+            postTracker.posts.append(contentsOf: response.posts)
+            postTracker.page += 1
+        }
+    } catch {
+        // appState.alertTitle = "Couldn't connect to Lemmy"
```
ShadowJonathan, line 45: Leftover?

Mlem/API/Requests/CreateCommentLikeRequest.swift
```diff
@@ -0,0 +1,49 @@
...
+        score: ScoringOperation
+    ) throws {
+        do {
+            let data = try JSONEncoder().encode(
+                Body(
+                    auth: account.accessToken,
+                    comment_id: commentId,
+                    score: score.rawValue
+                )
+            )
```
ShadowJonathan, line 35: I'm not entirely sure if I like the boilerplate of having to put the encoder instantiation and .encode function inside every .init every time, plus having to wrap everything in a try-catch, couldn't we just chuck a Body at an extension function or something, and have that deal with the encoding issue?

Mlem/API/Requests/CreateCommentLikeRequest.swift
```diff
@@ -0,0 +1,49 @@
...
+            self.endpoint = account.instanceLink
+                .appending(path: "comment")
+                .appending(path: "like")
```
ShadowJonathan, line 38: I'm also not sure if this is the most idiomatic way of declaring an API's path

Is there a way to define associated constants per structs? This way, the protocol just has to state there has to be a path string, which then gets added onto the instance via this, in an extension method or something, reducing more boilerplate

ShadowJonathan (changes requested):

Mlem/API/Models/Community/APICommunityModeratorView.swift
```diff
@@ -0,0 +1,17 @@
...
+
+import Foundation
+
+struct APICommunityModeratorView: Decodable {
+    let community: APICommunity
+    let moderator: APIPerson
+}
+
+extension APICommunityModeratorView: Identifiable {
+    var id: Int { moderator.id }
```
ShadowJonathan, line 16: It's a tuple of community and moderator, so this'll incorrectly equate when you're comparing with a different community

Mlem/API/Models/Posts/APIPost.swift
```diff
@@ -0,0 +1,38 @@
...
+    let id: Int
+    let languageId: Int
+    let local: Bool
+    let locked: Bool
+    let name: String
+    let nsfw: Bool
+    let published: Date
+    let removed: Bool
+    let thumbnailUrl: URL?
+    let updated: String?
```
ShadowJonathan, line 30: Is this not a date?

Mlem/Logic/Participating/Post Comment.swift
```diff
@@ -8,120 +8,53 @@
...
-            throw CommentPostingFailure.receivedInvalidServerResponse
-        }
-    }
-    catch let commentPostingError
-    {
-        print("Failed while posting comment: \(commentPostingError)")
-
-        throw CommentPostingFailure.couldNotPost
+func postComment(
+    to post: APIPostView,
```
ShadowJonathan, line 13: Why not just a post ID? (Same counts for a lot of these spots btw)

Maybe this is me being too nit picky, and this can be addressed later, but I'm noticing it here.

Mlem/Logic/Participating/Post Comment.swift
```diff
@@ -8,120 +8,53 @@
...
+    commentContents: String,
+    commentTracker: CommentTracker,
+    account: SavedAccount,
+    appState: AppState
+) async throws {
     let dominantLanguage = NSLinguisticTagger.dominantLanguage(for: commentContents)
+    let request = try CreateCommentRequest(
+        account: account,
+        content: commentContents,
+        languageId: dominantLanguage == "en" ? 37 : nil,
```
ShadowJonathan, line 47: We should really build up the list of languages... 😅

buresdv:
I'm leaving this entire PR in the very capable hands of @ShadowJonathan because this is way over my head 😅

Mlem/API/APIClient.swift
mormaer: Yeah, I agree, unfortunately if we `throw` then it'd take down the whole model with it, not just this property 😞 

I think the two options we have are;

1. make all `Date` properties optional by default, so in the above case we can return `nil` if we cannot decode it. This does mean the UI layer needs to handle that optionality, but would let the modelling succeed if we encounter rogue dates.
2. Fail (by throwing) when we cannot decode the date, as stated above however that would cause the entire modelling process to fail, so a bad date nested multiple layers deep in a response would fail... I'd like to avoid that.

I know you've been looking through the Rust source, is there a definitive list of date formats that the API could return? Perhaps if we ensure our formatter supports all possible date formats, but we still allow option 1 on failure it gives us the best solution... as we can be (mostly) confident all dates will format, but in an edge case we allow it to be omitted?

Mlem/API/APIRequest.swift
mormaer: This is just a weird one with how protocols are defined in Swift, you cannot declare a `let` inside a protocol definition, you instead declare if the protocol requires to read `{ get }`, or read and write `{ get set }`.

So in this case we need to use the keyword `var`, but declare conformance only requires read `{ get }`, inside the actual concrete structs they are still represented as `let` this way

Mlem/Logic/Networking/Infinite Loader.swift
mormaer: yeah, should be safe to remove this now 👍

Mlem/API/Requests/CreateCommentLikeRequest.swift
mormaer: makes sense - I'll take a look at that 👍

Mlem/API/APIClient.swift
ShadowJonathan: There should be a definitive list, which is why I suggest throwing, because rust wouldn't allow any other variant when we have looked at all of the ones that rust could possibly return.

Mlem/API/APIRequest.swift
ShadowJonathan: Ah, in that case this should be fine

I dont know much about swift protocols, thus the suggestion

Mlem/API/Requests/CreateCommentLikeRequest.swift
mormaer: we could adjust the protocol to require a path string, and then move the constructing into a default implementation - that should help, I'll take a look

Mlem/API/Models/Community/APICommunityModeratorView.swift
mormaer: ah! good catch, thanks, I'll update

Mlem/API/Models/Posts/APIPost.swift
ShadowJonathan: Just checked, this is an optional date:

```
  pub published: chrono::NaiveDateTime,
  pub updated: Option<chrono::NaiveDateTime>,
```

mormaer: it is, yeah - the front-end doesn't use this property yet so it's just modelled based on the type returned, I'll change those fields to be correct here and on the other models

Mlem/Logic/Participating/Post Comment.swift
mormaer: While refactoring I was trying not to adjust the existing methods too much, so previously this accepted the `Post` type, I just moved it to the new model... with the changes included in the refactoring quite a few of these arguments can be simplified and in other cases I think whole methods can be dropped as they mostly just take the arguments and build the request, which could now be done in-line...

To avoid (even more) scope creep I'd suggest we address this additional tidying up as a separate PR?

Mlem/API/APIClient.swift
ShadowJonathan: Lemmy uses `chrono`, a rust library, and specific a type which says its `ISO 8601`.

Mlem/Logic/Participating/Post Comment.swift
mormaer: whole heartedly agree - I'd suggest this be treated as a separate task, this one was already in the implementation so I kept it, but mapping between the API value and Apple's values would ideally be moved into an extension or such so we can use it throughout when required

mormaer:
> I'm leaving this entire PR in the very capable hands of @ShadowJonathan because this is way over my head 😅

@buresdv - once we've addressed the issues raised and feel it's in a good place, I'd really appreciate you doing some regression testing on the branch to make sure all the behaviours are correct if you have the time 🙏

Mlem/API/APIClient.swift
ShadowJonathan: We're not parsing the "Display" variant here, I think it'll be crash-less if we implement that.

Add a throw therafter, we can recover these crashes from testflight, or via the app store's auto-reporting.

Mlem/Logic/Participating/Post Comment.swift
ShadowJonathan: Maybe add a todo comment, so we know to return to it.

Mlem/Logic/Participating/Post Comment.swift
ShadowJonathan: Sounds good to me, was just commenting on the oddness

Mlem/API/APIClient.swift
mormaer: if I'm understanding correctly that suggests the two date formats we currently attempt should be the only two returned, basically the fractional seconds may or may not be present.

Adding a `throw` will break modelling so a single rouge date in a list of posts would mean the user cannot decode/see _any_ of their posts, the app wouldn't crash in response to the `throw`, we could **make** it crash but...

I'm wondering if this is actually a problem we should solve if/when we go to the repository pattern on top of this?

We allow optionality for now, which is the safest approach in terms of modelling the API and then if we later represent these with internal app models we could force non-optionality, this would mean a single broken post etc would simply not reach the UI layer - if a user reported such a missing post we could investigate, but it sounds like the API would be the one at fault should it ever happen?

buresdv:
>  I'd really appreciate you doing some regression testing on the branch to make sure all the behaviours are correct if you have the time 🙏

No problem at all, I will also enlist some additional testers from Mastodon

Mlem/API/APIClient.swift
ShadowJonathan: > we could make it crash but...

Please do, if it does give us crash logs, we can make it an exception that will show the date string it crashed on

I think that if we give back a nonsensical date, we will break the app in many more subtle ways, so i think thats a bad idea

ShadowJonathan: I dont want optionality, we have these models and we should stick with them

ShadowJonathan: from the dev room:

Jo:
> actually, there may be a better way to handle this, if we dont want to crash, and if we 
>
> dont want to throw or show an error
>
> what if we return Unix timestamp 0?
> 
> Jan 1 1970 is fairly well known for that kind of stuff

Mormaer:
> yeah, I say let’s throw for now, as it is most correct 👍
> 
> if we get hit with 100s of reports after that version is out we go with the reference date (1970) and consider it more
> 
> End of the day, us not being able to parse it (assuming the API is behaving) is an edge case

Mlem/API/APIClient.swift
ShadowJonathan: I was thinking for POST/PUT to accept any Data, but maybe then GET would accept a `[String: String]` then instead?

Mlem/Logic/Time Parser.swift
mormaer: addressed in 60d6229fce1469f7e2ee636f593e55e7796daf8c

Mlem/Logic/Networking/Infinite Loader.swift
mormaer: addressed in 60d6229fce1469f7e2ee636f593e55e7796daf8c

Mlem/API/APIClient.swift
mormaer: addressed by dbfe9740835ee1db247427226641e3617925b840

Mlem/API/Requests/CreateCommentLikeRequest.swift
mormaer: addressed by dbfe9740835ee1db247427226641e3617925b840

Mlem/API/Requests/CreateCommentLikeRequest.swift
mormaer: addressed by dbfe9740835ee1db247427226641e3617925b840

Mlem/Logic/Participating/Post Comment.swift
mormaer: added in a74f57900e668f3850afdfbcadd0a08a24dcb202

Mlem/API/Models/Posts/APIPost.swift
mormaer: addressed in 77b1f8cd998bd6c3d4209c7380cff943f4ea7932

Mlem/API/Models/Community/APICommunityModeratorView.swift
mormaer: fixed in 5cf9c1ef79e45e6b56bfcecc79afeb9edb8ab916

Mlem/API/APIClient.swift
mormaer: fixed in 8375791fbe626f0d5223919fdef9f474662bdabd

Mlem/Views/Tabs/Posts/Components/Post/Components/Instance and Community List/Add Account View.swift
```diff
@@ -235,13 +235,13 @@ struct AddSavedInstanceView: View
                 switch error {
                 case EndpointDiscoveryError.couldNotFindAnyCorrectEndpoints:
                     message = "Could not connect to \(instanceLink)"
-                case APIRequestError.encoding:
-                    message = "Invalid credentials"
                 case UserIDRetrievalError.couldNotFetchUserInformation:
                     message = "Mlem couldn't fetch you account's information.\nFile a bug report."
+                case APIClientError.encoding:
+                    message = "Invalid credentials"
```
ShadowJonathan, line 241: Why do we return this on an encoding error? Why don't we look at the 403 HTTP status code?

Mlem/Extensions/JSONDecoder+Default.swift
```diff
@@ -0,0 +1,41 @@
...
+            
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSS"
+            if let date = formatter.date(from: string) {
+                return date
+            }
+            
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+            if let date = formatter.date(from: string) {
+                return date
+            }
```
ShadowJonathan, line 29: You're missing the not-T variant as described [here](https://github.com/buresdv/Mlem/pull/95#discussion_r1225766951), otherwise its good

Mlem/API/APIClient.swift
```diff
@@ -91,4 +51,31 @@ class APIClient {
...
+        defintion.headers.forEach { header in
+            urlRequest.setValue(header.value, forHTTPHeaderField: header.key)
+        }
+        
+        if let _ = defintion as? any APIGetRequest {
+            urlRequest.httpMethod = "GET"
+        } else if let postDefinition = defintion as? any APIPostRequest {
+            urlRequest.httpMethod = "POST"
+            urlRequest.httpBody = try createBodyData(for: postDefinition)
+        } else if let putDefinition = defintion as? any APIPutRequest {
```
ShadowJonathan, line 66: `APIPutRequest` is defined as a `APIPostRequest` subclass, wont the above selection (`as? PostRequest`) catch all put requests?

ShadowJonathan, line 66: `APIPutRequest` is defined as a `APIPostRequest` subclass, wont the above selection (`as? PostRequest`) catch all put requests?

ShadowJonathan (commented):
Looking good so far

Mlem/Views/Tabs/Posts/Components/Post/Components/Instance and Community List/Add Account View.swift
mormaer: Ah, this is perhaps just bad wording on my part, an`.encoding` error means we failed to parse to body of the request _before_ sending... we should raise in the creative channel that it'd be worth developing a _tone_ for messaging in the app so we can all agree on copy for different error scenarios... any thoughts on this one?

Something like:

> Please check your username and password

ShadowJonathan: I think that's a better message, yeah.

And if this is before sending, why are we checking it here? Why are we not checking the form elsewhere?

I think this should at least get a TODO to abstract out more into the UX side of things, so we don't get partial values.

mormaer: yeah, the inputs need validatio/sanitised as currently they're not - but I agree that's a job for another PR, I'll add a TODO 👍

mormaer: d37f4ea4e109ff48bab2c41ecee5c57c9a3d76a8

Mlem/API/APIClient.swift
mormaer: oh my yes 🤦 - it's been a long day -> 3fe6eeef30d4b2c69fff9a69aecdf4360ba684ba

fixed by not doing protocol composition, but having them both inherit a protocol that describes having _bodies_

Mlem/Extensions/JSONDecoder+Default.swift
mormaer: I think this (b6ca0ec6a8d80bf3a3f47a200393c75d8d6f744c) is what you mean? I may be mistaken? 😅

ShadowJonathan: Yes, this is good

Mlem/API/APIClient.swift
ShadowJonathan: Somehow this comment duplicated? huh

ShadowJonathan (approved):
Looking good

PR #97, opened 2023-06-11
Add 2fa token text field to login

justincmoy:
# Checklist

# Pull Request Information

## About this Pull Request

Following the (eventual [0],[1],[2]) addition of a totp token to the login workflow, this adds a text field upon receiving a specific error message from the server.

[0] https://github.com/LemmyNet/lemmy/pull/2741
[1] https://github.com/LemmyNet/lemmy-js-client/pull/106
[2] https://github.com/LemmyNet/lemmy-ui/commit/af1e8868bd13df441ea02f9c13f5c3ee359db597

## Screenshots and Videos

The following video includes:
- Showing and focusing the 2FA token text field following login that requires the token, without showing the error returned by the server.
- Showing the error message returned by the server if the 2FA token is incorrect.
- Successfully logging in as a user with a correct 2FA token.

https://github.com/buresdv/Mlem/assets/14875982/60b9c9bb-2ac3-4382-938a-92693075f164

## Additional Context


mormaer:
Nice, we should definitely get this supported and should be simple enough to integrate with the changes in #95 👍 

There is some basic error handling in the other PR which would catch that error, but currently it just wraps the error message from the API (`"missing_totp_token"`) and returns it for handling... I agree that it should likely return it's own bespoke error for this case.

It's unfortunate that the error model returned is just a `String`, as determining behaviour using a string comparison is less than ideal, but the API is the API and for now that may be the best option we have 😞

buresdv:
This is a great addition. I believe that the 2fa field should be hidden by default and only show up if the server responds with the "missing token" message. That way, users that don’t have it configured don’t have to leave an empty field. Users expect to have to fill in every available field.

justincmoy:
> It's unfortunate that the error model returned is just a `String`, as determining behaviour using a string comparison is less than ideal, but the API is the API and for now that may be the best option we have 😞

I agree. If anything, this PR could do a better job as the `missing_totp_token` String is a value of the `.error` JSON key, so perhaps we could attempt to unmarshal the response body to json, and then switch-case on the error key.

> I believe that the 2fa field should be hidden by default and only show up if the server responds with the "missing token" message.

That's how this PR currently behaves via the `isShowingTwoFactorSection` flag.

buresdv:
> That's how this PR currently behaves via the `isShowingTwoFactorSection` flag.

Great, I didn't notice that. Maybe one more suggestion would be to not show the "Please enter the 2FA token" as an error (because some users might get confused by us showing an error while it's not really an error, just a request for more information) and instead just reveal and focus the 2FA field?

ShadowJonathan:
What if you show the "please enter the 2FA token" as a popup with an entry field?

justincmoy:
> > That's how this PR currently behaves via the `isShowingTwoFactorSection` flag.
> 
> Great, I didn't notice that. Maybe one more suggestion would be to not show the "Please enter the 2FA token" as an error (because some users might get confused by us showing an error while it's not really an error, just a request for more information) and instead just reveal and focus the 2FA field?

Done!

Mlem/Views/Tabs/Posts/Components/Post/Components/Instance and Community List/Add Account View.swift
```diff
@@ -137,6 +143,24 @@ struct AddSavedInstanceView: View
...
+                    if isShowingTwoFactorText {
+                        HStack
+                        {
+                            Text("2FA Token")
+                            Spacer()
+                            SecureField("TwoFactorToken", text: $twoFactorToken, prompt: Text("000000"))
+                                .focused($focusedField, equals: .twoFactorField)
+                                .submitLabel(.go)
+                                .keyboardType(.asciiCapableNumberPad)
+                                .onAppear
```
WestonHanners, line 156: `.textContentType(.oneTimeCode)`
https://developer.apple.com/documentation/uikit/uitextcontenttype/2980930-onetimecode

mormaer (approved):
All good from me, the `+4` second delay hurts my soul but until we can refactor this view it's a necessity - happy for this to be merged if you are @WestonHanners, we could always add the accessibility modifier for one time code as a follow-up?

Issue #52, opened 2023-06-08
Write internal API Client


ShadowJonathan:
This has been corroborated with the matrix room; I'd like to rewrite the internal API Client (the functions and data models) that calls out to the Lemmy servers.

I'll put the application Logic under `Logic/Networking/Lemmy Interface`, and the data models under `Models/Lemmy Interface`.

Doing this correctly is a bit of a challenge since Lemmy currently does not have HTTP API Docs, but ill reverse engineer the API from the rust source.

ShadowJonathan:
The relevant API code is here: https://github.com/LemmyNet/lemmy/blob/0.17.3/src/api_routes_http.rs

ShadowJonathan:
cc @mormaer

I've extracted all of the endpoints from Lemmy with their corresponding HTTP method and associated rust type, which then links to the actual performance code, and the return type

<details>
<summary>A lot of endpoints</summary>

```
GET /site GetSite
POST /site CreateSite
PUT /site EditSite

GET /modlog GetModlog

GET /search Search

GET /resolve_object ResolveObject

GET /community GetCommunity
POST /community CreateCommunity
PUT /community EditCommunity
PUT /community/hide HideCommunity
GET /community/list ListCommunities
POST /community/follow FollowCommunity
POST /community/block BlockCommunity
POST /community/delete DeleteCommunity
POST /community/remove RemoveCommunity
POST /community/transfer TransferCommunity
POST /community/ban_user BanFromCommunity
POST /community/mod AddModToCommunity
POST /community/join CommunityJoin
POST /community/mod/join ModJoin

GET /post GetPost
POST /post CreatePost
PUT /post EditPost
POST /post/delete DeletePost
POST /post/remove RemovePost
POST /post/mark_as_read MarkPostAsRead
POST /post/lock LockPost
POST /post/feature FeaturePost
GET /post/list GetPosts
POST /post/like CreatePostLike
PUT /post/save SavePost
POST /post/join PostJoin
POST /post/report CreatePostReport
PUT /post/report/resolve ResolvePostReport
GET /post/report/list ListPostReports
GET /post/site_metadata GetSiteMetadata

GET /comment GetComment
POST /comment CreateComment
PUT /comment EditComment
POST /comment/delete DeleteComment
POST /comment/remove RemoveComment
POST /comment/mark_as_read MarkCommentReplyAsRead
POST /comment/like CreateCommentLike
PUT /comment/save SaveComment
GET /comment/list GetComments
[...]

PR #120, opened 2023-06-12
Go over the current API and sync it to Lemmy's types

ShadowJonathan:
Related to https://github.com/buresdv/Mlem/issues/52

# Checklist

# Pull Request Information

## About this Pull Request
This PR refactors most of the existing types to use the ordering of the fields as done on Lemmy, to simplify comparisons.

This also fixes some types and dates, and stubs out existing types with more fields.


buresdv (dismissed):
yet another great find

buresdv:
There seem to be some conflicts, can you have a look, please? I'm not on the computer atm so I can't do much.

Mlem/API/Requests/Comment/GetCommentsRequest.swift
```diff
@@ -0,0 +1,55 @@
...
+        savedOnly: Bool? = nil
+    ) {
+        self.instanceURL = account.instanceLink
+        self.queryItems = [
+            .init(name: "auth", value: account.accessToken),
+            .init(name: "post_id", value: "\(postId)"),
+            .init(name: "max_depth", value: "\(maxDepth)"),
+            .init(name: "type_", value: type.rawValue),
+            .init(name: "sort", value: sort),
+            .init(name: "page", value: page.map(String.init)),
```
mormaer, line 40: why for some of the values do we need to add the `.map`, but for other optional arguments like `sort` above we don't?

some of these arguments were not present before, which suggests they can be optional and not sent, if that were the case given the filtering done later, could they not all be simplified to just:
```swift
.init(name: "name", value: value)
```

mormaer: oh hang on, I'm being silly - it's because they all need coerced into `String` for the types which aren't 🙈 

carry on

mormaer (approved):
looks good - one small suggestion, would adding a link to the schema location in the Lemmy repo alongside the schema names be valuable?

ShadowJonathan:
> looks good - one small suggestion, would adding a link to the schema location in the Lemmy repo alongside the schema names be valuable?

That's essentially what the schema names are, adding links would imo become reundant or possibly outdated, and are a lot more verbose

Mlem/API/Requests/Post/GetPostsRequest.swift
```diff
@@ -29,23 +33,36 @@ struct GetPostsRequest: APIGetRequest {
...
-        
+
         if let communityId {
             queryItems.append(
                 .init(name: "community_id", value: "\(communityId)")
             )
         }
-        
+
+        if let limit {
```
mormaer: for consistence should these also be changed to use the `.map` approach?

Mlem/API/Requests/SearchRequest.swift
```diff
@@ -17,28 +18,65 @@ enum SearchType: String, Codable {
...
-        self.queryItems = [
+        var queryItems: [URLQueryItem] = [
             .init(name: "auth", value: account.accessToken),
             .init(name: "type_", value: searchType.rawValue),
             .init(name: "sort", value: sortOption.rawValue),
             .init(name: "listing_type", value: listingType.rawValue),
             .init(name: "q", value: query)
         ]
+
+        if let page {
```
mormaer: as above ^

mormaer (approved):

Issue #29, opened 2023-06-04
Replace navigation with NavigationStack


buresdv:
As the old `NavigationView` is deprecated in iOS 16, it should be migrated to use the new `NavigationStack` API.

Ideally, the NavigationStack would make use of a path so arbitrary screens can be pushed onto the stack (like the sidebar, which at the moment uses the old `isActive` method of pushing it to the stack).

## Discussion
There are some challenges, namely
- The need to use the `path:` property on `NavigationStack`, so there needs to be an implementation for a tracker that will store the path and modify it. The path needs to be set so that we can push `CommunitySidebarView` onto the stack when the user clicks on the "Sidebar" button in me menu.
- #30 could be related to the old implementation of `NavigationView`. Maybe converting to `NAaigationStack` would fix it.

buresdv:
Done in a fast way by replacing `NavigationView` definition with `NavigationStack`. Nevertheless, that fixed #30.

Implement a more detailed implementation later

tht7:
Stale! navigation was replaced and simplified already

PR #143, opened 2023-06-14
Fix #68 - crash when clicking community links

### LINKED ISSUE #68: crashing when clicking Universal Links
**Describe the bug**
clicked link of the format [News](/c/news@beehaw.org)

**Expected Behavior**
What did you expect to happen?
open news@beehaw.org

**Actual Behavior**
What actually happened?
app crashed

**To Reproduce**
What steps do I need to take to trigger the bug?
click a link of the above format they are easy to find in (https://lemmy.world/c/newcommunities)

**Mlem Information:**
Version + Build number
0.0.6 (21)
If this started happening recently, write which version introduced the bug
unknown

**Additional Context**
Any other relevant information
I dont know if it will happen will all links but it has happend with every link i have tried

mormaer:
# Checklist

# Pull Request Information

## About this Pull Request
This PR addresses the crash described in #68. The application was crashing as links in the style `!abc.def.ghi` are picked up by the `MarkdownView` as `mailto:abc@def.ghi`, we'd then attempt to feed this directly into Safari which Apple don't like 🙈.

As a first step this PR adds a small `guard` to ensure the URL is of the http/https protocol before proceeding with pushing Safari, any links that are not in that format will be `.discarded` and a generic alert will be presented to the user.

## Next steps

This PR resolves the crash however we want to get to a place where links are either handled natively in the application, or handed out to the OS/Safari.

The plan is to address #29 next to move to the new navigation APIs, and then #144 to correctly recognise links.

After which we should be unblocked for implementing deep linking by using the `resolve_object` API call to retrieve the corresponding community/post/user objects we need to display the content within the application.

## Considerations

One thing to note is that genuine `mailto:` links will now also display this alert for now, but I don't consider this a _regression_ as they would have crashed the application before this change. 🤷 

Another approach we could have would be that instead of discarding unsupported links we return `.systemAction`. This would open the users email client should they click an email link, but due to links in the format `!abc@def.ghi` being incorrectly parsed as `mailto:` currently we'd do the same 😞.

At present I think considering them unsupported is the best approach in lieu of the future work to correctly handle all links.

## Screenshots and Videos
| ALERT |
| --- |
| alert |

## Additional Context

### DIFF
--- a/Mlem/Logic/URLHandler.swift
+++ b/Mlem/Logic/URLHandler.swift
@@ -29,6 +29,7 @@ class URLHandler {
     /// - Returns: A `Result` containing to the system level `OpenURLAction.Result` and any application level actions to perform
     static func handle(_ url: URL) -> Result {
         guard let scheme = url.scheme, scheme.hasPrefix("http") else {
+            // TODO: handle additional link types appropriately
             // as the current handling only supports http(s) via Safari bail early if the scheme is unsupported...
             // a future piece of work will add deep linking where possible with help of the `ResolveObject` API call
             return .init(result: .discarded, action: .error("This type of link is not currently supported 😞"))
--- a/Mlem/Logic/URLHandler.swift
+++ b/Mlem/Logic/URLHandler.swift
@@ -15,7 +15,12 @@ class URLHandler {
     init() { /* class should not be instantiated */ }
     
     static func handle(_ url: URL) -> OpenURLAction.Result {
-        #warning("TODO: consider how we might deep link within the application for urls such as '/c/<community_name>' and '/post/<post_id>'")
+        guard let scheme = url.scheme, scheme.hasPrefix("http") else {
+            // as the current handling only supports http(s) via Safari bail early if the scheme is unsupported...
+            // a future piece of work will add deep linking where posssible with help of the ResolveObject API call
+            return .discarded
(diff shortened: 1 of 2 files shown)


Mlem/Logic/URLHandler.swift
```diff
@@ -11,14 +11,33 @@ import SafariServices
...
     
-    static func handle(_ url: URL) -> OpenURLAction.Result {
-        #warning("TODO: consider how we might deep link within the application for urls such as '/c/<community_name>' and '/post/<post_id>'")
+    /// A method to perform handling on URLs within the application
+    /// - Parameter url: The `URL` you require to be handled
+    /// - Returns: A `Result` containing to the system level `OpenURLAction.Result` and any application level actions to perform
+    static func handle(_ url: URL) -> Result {
+        guard let scheme = url.scheme, scheme.hasPrefix("http") else {
+            // as the current handling only supports http(s) via Safari bail early if the scheme is unsupported...
+            // a future piece of work will add deep linking where possible with help of the `ResolveObject` API call
```
ShadowJonathan, line 33: Maybe add TODO here?

ShadowJonathan (approved):
im not entirely versed on the UI codestyle/patterns side of things for this, but LGTM

PR #145, opened 2023-06-14
Add a lot of API types

ShadowJonathan:
This adds a lot of types from #52, the only missing ones are a lot of the PM, User, Admin, and the Misc endpoints


Mlem/API/Requests/Site/GetSite.swift
```diff
@@ -0,0 +1,42 @@
...
+        self.instanceURL = instanceURL
+
+        self.queryItems = []
+    }
+}
+
+// lemmy_api_common::site::SiteResponse
+struct SiteResponse: Decodable {
+    let site_view: APISiteView
+    // TODO
```
ShadowJonathan, line 41: d'oh 🤦

Mlem/API/Requests/ResolveObject.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    init(
+        account: SavedAccount,
+
+        query: String
+    ) {
+        self.instanceURL = account.instanceLink
+        self.queryItems = [
+            .init(name: "q", value: query),
+
```
mormaer, line 27: would you object to removing these blank lines here and in the others?

in a request the size of `EditSite` I think the occasional break works to help split out the sections, but in many of the others like here it just looks off?

I'm thinking it's designed to separate the auth/account argument from the other fields, but with so few it doesn't feel necessary?

Mlem/API/Requests/Site/EditSite.swift
```diff
@@ -0,0 +1,165 @@
...
+    let body: Body
+
+    // lemmy_api_common::site::CreateSite
+    struct Body: Encodable {
+        let name: String?
+        let sidebar: String?
+        let description: String?
+        let icon: URL?
+        let banner: URL?
+        let enableDownvotes: Bool?
```
mormaer, line 25: you're about to hate me... but I just realised that all these `Body` structs are using `lowerCamelCase` for their fields, if you look in the previously created models I was using `snake_case` so that what we encode and send to the API is correct:
```swift
// lemmy_api_common::post::CreatePostLike
struct Body: Encodable {
    let auth: String
    let post_id: Int
    let score: Int
}
```

The encoder in the API is set to use the default key encoding strategy so will not convert these new ones, we could enable that (and update all the existing models) but given they're internal models only used within the request initialisers I'd gone the path of least resistance... and although likely not a consideration for the size of bodies we're encoding Apple state the following:

> Key encoding strategies other than JSONEncoder.KeyEncodingStrategy.useDefaultKeys may have a noticeable performance cost because those strategies may inspect and transform each key.

I just checked by changing the existing `CreatePostLike` call to use `postId` in it's body instead of `post_id` and the API doesn't like that and spits back:

> "Json deserialize error: missing field `post_id` at line 1 column 182"

mormaer (changes requested):

Mlem/API/Requests/Comment/GetComment.swift
```diff
@@ -0,0 +1,43 @@
+//
+//  X.swift
```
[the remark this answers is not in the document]
ShadowJonathan: Oops

Mlem/API/Models/Posts/APIPostReportView.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  APIPostReportView.swift
+//  Mlem
+//
+//  Created by Nicholas Lawson on 09/06/2023.
```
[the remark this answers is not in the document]
ShadowJonathan: Copy paste error 😅

Mlem/API/Models/Site/APIMyUserInfo.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Mlem
+//
+//  Created by Jonathan de Jong on 14.06.2023
+//
+
+import Foundation
+
+// lemmy_api_common::site::MyUserInfo
+struct APIMyUserInfo: Decodable {
+    // TODO
```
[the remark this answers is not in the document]
ShadowJonathan: Wanted to do that in a follow up PR, because this structure gets a little expansive

Mlem/API/Requests/Site/EditSite.swift
ShadowJonathan: image

(Jk, it's no problem, thanks for catching this 💚)

Mlem/API/Requests/ResolveObject.swift
ShadowJonathan: > would you object to removing these blank lines here and in the others?

they're mainly meant for me to easily insert and discern the properties that're not `auth`, I think it works for code readability like that.

Mlem/API/Requests/Post/CreatePost.swift
```diff
@@ -32,6 +33,7 @@ struct CreatePostRequest: APIPostRequest {
         name: String,
         nsfw: Bool?,
         body: String?,
+        // TODO change to `URL?`
```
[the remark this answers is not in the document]
ShadowJonathan: Hm? What do you mean?

This is more because im not entirely sure if they'd be valid `URL`s, but I added a todo here to check later.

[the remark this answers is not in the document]
mormaer: ah fair enough, I just noticed I've added a TODO further down in the same file because what we get passed isn't sanitised, so but to leave it as it'd require work at the UI layer to send a guaranteed one in 👍

Mlem/API/Requests/ResolveObject.swift
mormaer: no worries, personally I'd always put the auth argument first, but as developers we can debate coding style until the heat death of the universe and never reach consensus 🙃 

On a related note, I was planning to bring [swiftformat](https://github.com/nicklockwood/SwiftFormat) into the project when I get a moment, likely set up as a pre-commit hook, what are your thoughts?

The thinking is it'll ensure the project is formatted to the _"official"_ language style, and saves any style based stuff creeping into reviews?

Mlem/API/Models/Site/APILocalSite.swift
```diff
@@ -0,0 +1,45 @@
...
+    let defaultPostListingType: String
+    let legalInformation: String?
+    let hideModlogModNames: Bool
+    let applicationEmailAdmins: Bool
+    let slurFilterRegex: String?
+    let actorMaxNameLength: Int
+    let federationEnabled: Bool
+    let federationDebug: Bool
+    let federationWorkerCount: Int
+    let captchaEnabled: Bool
```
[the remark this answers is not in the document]
ShadowJonathan: I haven't implemented registration yet, there's captchas for endpoints, yeah

Mlem/API/Requests/Site/EditSite.swift
ShadowJonathan: This is done now

Mlem/API/Requests/ResolveObject.swift
ShadowJonathan: I'll make an issue about this, so we can introduce it later

ShadowJonathan:
I'm converting this to draft, so that I can implement the rest of the API types

PR #133, opened 2023-06-12
Post Overhaul

EricBAndrews:
# Checklist

- [x] I have described what this PR contains

**Choose one of the following two options:**
- - [ ] This PR does not introduce major changes
- - [x] This PR introduces major changes, and I have consulted @buresdv, @jboi or @mormaer in the *Mlem Development Matrix room*

**Choose one of the following two options:**
- - [ ] This PR does not change the UI in any way
- - [x] This PR adds new UI elements / changes the UI, and I have attached pictures or videos of the new / changed UI elements

# Pull Request Information

## About this Pull Request
This PR overhauls the look of posts to align with Figma, adds compact posts, and refactors the underlying post views.

### Compact posts

Posts can now be displayed compactly in the feed. This option can be turned on from settings.

### Instant upvote feedback

Upvotes and downvotes now display immediately upon the user tapping the button. If the server fails to register the upvote, the visual indicator is removed when that `async` call completes.

### Upvote and Downvote Button Restyling

The vote buttons' status is now displayed by their backgrounds, making it easier to read and colorblind-friendly. 

### Overflow options

Additional options (right now, only share post) have been moved to an overflow menu. It is accessible via the ellipsis button on a large post or by long pressing a compact post.

### New view structure

`PostItem` has been refactored into `FeedPost` and `ExpandedPost`, where `FeedPost` renders posts in the `CommunityView` feed and `ExpandedPost` renders posts in the detail view.

Supporting these two views are the following subviews:
- `LargePost` displays the post author and post body (full image for image posts, WebComplex for links, and full or truncated text text posts depending on the state of the `isExpanded` bool). It is used both to display posts in the feed and in the detail view.
- `CompactPost` displays a compact title card for the post (see screenshots)
- `PostInteractionBar` contains all of the core post interactions: upvote, downvote, save, reply. It can render both normally or compacted, with the compacted view being a little smaller and lacking the divider. 
*Design note: it doesn't* exactly *match the Figma. I have put reply and save on the right and the post info in the middle; the motivation for doing this is to declutter the interaction space and visually cue the user into the (planned) swipe controls--upvote and downvote would come in from the left, while reply and save would come in from the right. Since it's all grouped into one view, this design can easily be changed if it is deemed unsuitable.*
- `PostHeader` is a helper view for `LargePost`.

### New post views

<p float="left">
New large post view
Compact post view with upvote
</p>

### New share menu

<p float="left">
Compact mode share contextmenu
Share contextmenu
</p>

### Compact mode menu option

[...]

Issue #117, opened 2023-06-12
Add abstraction for upvoting/downvoting "in advance"


ShadowJonathan:
Inspired by https://github.com/buresdv/Mlem/issues/115, I'd like for there to be a struct that'll hold the "actual count" together with the "virtual count", the "virtual" count being updated with user action, and then it kicking off a process to update the API to reflect this status.

This way, we don't need to fiddle with the API response to mock/trick the frontend into thinking it is the actual class, and also have a struct that will automatically debounce/adjust the voting status according to the user selection, and update the "upvoted" state back to the previous one when it fails, cascading through to the frontend.

ShadowJonathan:
An additional note: This approach should queue up additional votes when given, and/or just store them as the "desired final state".

Something akin to the following events;
```
- User downvote
- Downvote action is stored and executed
- User resets downvote
- Downvote action is still ongoing, reset action is queued
- User upvotes
- Downvote action is *still* ongoing, reset action is replaced with the upvote action
- Downvote action completes, upvote action is executed
- Upvote action finished
```

This way, successive calls (such as the reset first) is avoided, while the final vote is made "eventually consistent". We should also possibly ignore network/HTTP errors with the downvote action if there is a "more desired state" (such as the reset or upvote action).

All of these actions, such as storing/replacing the queued action, and "popping" it, should be done atomically, so that (for example) if the user would reset their vote again while the upvote action is being grabbed, it would either replace the upvote action, or would queue it after.

Issue #115, opened 2023-06-12
* Regressions in the new API release


buresdv:
**Describe the bug**
Here I will be cataloguing all the regressions I have encountered trying out the version with the new API

- [x] The Subscribed feed doesn't show Subscribed communities, instead it’s just the All feed
- [x] When voting on comments/posts, the “fake” instant upvote is no longer working. Previously, it would change the UI immediately and wait for the response in the background, but now it waits without updating the UI. This breaks the UX pretty bad.
- [x] #19 is back
- [x] After searching, the search results don’t get cleared. This also makes the "Cancel" button not show up, leading to this unrecoverable state, where you can't close the search: ![image](https://github.com/buresdv/Mlem/assets/22037369/eda81076-017e-4f46-915b-782b7b413288)

PR #156, opened 2023-06-15
Fix - instant comment interactions

mormaer:
# Checklist

- - [x] This PR does not change the UI in any way

# Pull Request Information

## About this Pull Request
This PR addresses the last issue described on https://github.com/buresdv/Mlem/issues/115.

After integrating the new `APIClient` we were no longer _faking_ the up/downvote interaction while waiting on the response which causes a delay in the UI updating regression the UX of this area.

The changes in this PR follow a similar approach to that implemented in #133 where a small piece of local state is introduced which holds the _expected_ new values if our voting operation succeeds. Once we have a response from the server we remove our local state so that we remain in sync.

These changes are related to but do not address https://github.com/buresdv/Mlem/issues/117. Once we have both this and the implementation for faking the post voting in `master`, we can look at how to bring them together under an abstraction as requested in that issue.

## Additional Context
Although this is only a partial fix while we work toward a unified approach across posts and comments it resolves the regression in the UX ahead of our next public release.

### DIFF
--- a/Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
+++ b/Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
@@ -21,12 +21,40 @@ struct CommentItem: View
     @State var isCollapsed = false
     
     @State private var isShowingTextSelectionSheet: Bool = false
+    @State private var localCommentScore: Int?
+    @State private var localVote: ScoringOperation?
+    
+    /// The color to use on the upvote button depending on our current state
+    private var upvoteColor: Color {
+        let vote = localVote ?? hierarchicalComment.commentView.myVote
+        
+        switch vote {
+        case .none, .downvote, .resetVote:
+            return .accentColor
+        case .upvote:
+            // TODO: when the posts overhaul merge is in this should use the same value
+            return .green
+        }
+    }
+    
+    /// The color to use on the downvote button depending on our current state
+    private var downvoteColor: Color {
+        let vote = localVote ?? hierarchicalComment.commentView.myVote
+        
+        switch vote {
+        case .none, .upvote, .resetVote:
+            return .accentColor
+        case .downvote:
+            // TODO: when the posts overhaul merge is in this should use the same value
+            return .red
+        }
+    }
 
     var body: some View
     {
         VStack(alignment: .leading, spacing: 10)
         {
-            if hierarchicalComment.commentView.comment.deleted == true
+            if hierarchicalComment.commentView.comment.deleted
             {
                 Text("Comment was deleted")
                     .italic()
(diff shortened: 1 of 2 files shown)


Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
```diff
@@ -195,19 +205,33 @@ struct CommentItem: View
     }
     
     private func rate(_ comment: HierarchicalComment, operation: ScoringOperation) async throws {
+        guard localVote == nil else {
+            // if we have a local vote then we're in the middle of rating
+            // so avoid the user being able to initiate additional requests
```
ShadowJonathan, line 198: This could result in a bug where an additional vote (such as a downvote after accidentally pressing up, or even resetting the vote) would be discarded, is there a way to queue this up, or something?

ShadowJonathan (changes requested):
This looks good, but can you also make this for the Post? Currently it only does this for comments, while it's still delayed for posts.

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
ShadowJonathan: Possibly a "deferredAction" that'll get replaced atomically when this function is called, and taken atomically at the end of the rating function, to prevent data races

mormaer:
> This looks good, but can you also make this for the Post? Currently it only does this for comments, while it's still delayed for posts.

It has been resolved for posts in #133 🤞

ShadowJonathan:
In that case, my objection is just to make sure follow-up votes get properly queued, and not discarded.

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
mormaer: This is true, however I think any approach like that would be best tackled as part of #117, so we can have a unified system that works across posts and comments... arguably it could even be an abstraction that allows for any type of deferred set of actions?

There is some thought needed as well as to what our state should be, as we'd need to predict the new state on each press, but defer the call till we know the outcome of the in-flight one, and on a patchy network or overloaded server, what if three calls queue-up and one fails? What if some blip caused them to return in a different order? How many should we queue, as the user could just keep tapping indefinitely and spam the API 😬 

My preference would be that we disable the interaction while a call is in-flight, there is a `.disabled(...)` modifier in `SwiftUI` which I initially added, but that flips the colour of the buttons to being greyed out which looked a bit... iffy, since the call is normally returned quickly it flicks between standard/greyed out so quickly it looked glitchy 😞 

In terms of the current UX with the early return/guard, there is a small haptic feedback triggered when the user taps on the button to indicate they've submitted a rating, this happens when the call initiates so in the above case no haptic feedback occurs which I think (for now) indicates to the user that the second/third/fourth presses are not registering.

Perhaps design could consider how best to communicate a call is in-flight, perhaps a small visual indicator is what we need to resolve the above without adding more complexity?

ShadowJonathan: I think we could ignore this then, and add it to the known bugs for the next release.

mormaer:
> In that case, my objection is just to make sure follow-up votes get properly queued, and not discarded. (The same counts for #133, if it has that bug)

I just seen your comment [here](https://github.com/buresdv/Mlem/pull/133#discussion_r1230868466), and yes the post side of things would indeed make repeated calls if you kept tapping, I'll stick a note over there asking if we can update it to also _ignore_ subsequent taps in the meantime

PR #173, opened 2023-06-16
Clarify that Mlem is source-available and not open-source

### LINKED ISSUE #170: Clarify license in the README
The README states that

https://github.com/buresdv/Mlem/blob/344fe0c91966fdf7ff465ccdaac6f7f1ebd8cdc5/README.md#L9

https://github.com/buresdv/Mlem/blob/344fe0c91966fdf7ff465ccdaac6f7f1ebd8cdc5/README.md#L48-L50

This is however not entirely correct, Commons Clause is a license that (when applied to an OSS license, probably a good idea to specify the base license too) makes a project [source-available](https://en.wikipedia.org/wiki/Source-available_software), but _not_ [open-source](https://en.wikipedia.org/wiki/Open-source_software), the latter has a pretty specific definition that e.g. [cannot forbid commercial use](https://opensource.stackexchange.com/questions/9805/can-i-license-my-project-with-an-open-source-license-but-disallow-commercial-use).

fwcd:
# Checklist

- [x] This PR does not introduce major changes
- [ ] This PR introduces major changes, and I have consulted @buresdv, @jboi or @mormaer in the *Mlem Development Matrix room*

- [x] This PR does not change the UI in any way
- [ ] This PR adds new UI elements / changes the UI, and I have attached pictures or videos of the new / changed UI elements

# Pull Request Information

## About this Pull Request
Fixes #170, since Commons Clause licenses are _not_ open-source as per the OSI definition, the project shouldn't advertise itself as such. See the FAQ on https://commonsclause.com/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Lemmy client for iOS.
 [![Mastodon Link](https://img.shields.io/mastodon/follow/108939255808776594?domain=https%3A%2F%2Fmstdn.social&label=Follow%20me%20for%20updates&style=flat)](https://elk.zone/mstdn.social/@davidbures)
 
 ## What is Mlem?
-Mlem is a client for [Lemmy](https://join-lemmy.org), a Reddit-esque, open-source link aggregator. With Mlem, you can effortlessly participate in the conversation across all Lemmy servers.
+Mlem is a client for [Lemmy](https://join-lemmy.org), a Reddit-esque, source-available link aggregator. With Mlem, you can effortlessly participate in the conversation across all Lemmy servers.
 
 ## Why Use Mlem?
 
@@ -47,4 +47,4 @@ As of now, Mlem is still in beta. While it already has many core features, there
 
 Mlem is licensed under [Commons Clause](https://commonsclause.com).
 
-This means that Mlem is open-source and you can do whatever you want with its source, like modifying it, contributing to it etc., but you can't sell Mlem or modified versions of it.
+This means that Mlem is source-available and you can do whatever you want with its source, like modifying it, contributing to it etc., but you can't sell Mlem or modified versions of it.


ShadowJonathan:
If an organisation defines "Open Source" as permitting essentially theft-of-effort, then we don't agree with this definition on those terms.

Open Source is something any organisation can change on any whim, its a movement and ideal that isn't beholden to any single word-of-definition.

Splitting hairs over this is counterproductive.

fwcd:
This is not about philosophy, it's about not making false claims.

I agree that there are valid reasons for restricting commercial use, but please don't call it "open source" when it violates the widely accepted definition which even the Commons Clause agrees on.

ShadowJonathan:
The Commons Clause is simply the best license for our usecase we've come across, but like I said, if the OSI tries to redefine Open Source as permitting universal commercialisation, then we don't agree with that as countering the spirit of Open Source.

We would've chosen CC BY-NC, were it not for the fact we also need to fund our own side of the project, such as servers (in the future, to likely handle notifications), or Apple's fees (developer subscriptions, and god knows what else).

If you know a better license that addresses this situation (We need to be able to commercialise it for our own costs, but we dont want company to rip it off, slap a new label on it, put some features behind paywalls, and then sell it behind our backs, without any regard for our effort), then we'll might consider that one.

Anyhow, we don't see how we're breaking any of the spirit of Open Source in the process, in fact, personally, I'd say that noncommercialisation was meant to be a core part of the beginnings of Open Source movement, of its roots, what now has evolved in the "Libre"/"Open" split definitions, and the fighting amongst them.

We don't want to partake in the hair-splitting, and we just want to work on a good app, calling it "source available" means we're just opting into this mess of a debate around definitions, which we really don't want to do.

It's like calling a man not a man because he doesn't have his balls.

ssddanbrown:
> if the OSI tries to redefine Open Source

Much of the open source world, and the popularity of the term, has been built upon that OSI definition.

>  ...were it not for the fact we also need to fund our own side of the project...
> ...We need to be able to commercialise it for our own costs...

The reality is that you're attempting to put your business interests first via the license, over the freedoms and use of users and the code. Nothing wrong with that IMO, and I respect the right to protect your efforts, but it's not in the spirit of open source.

>  I'd say that noncommercialisation was meant to be a core part of the beginnings of Open Source movement, of its roots

Not really. Both open source and the free software movement have required commercial use.
General commercial use is not really a threat to the free or open source idea, it's more about the fundamental freedoms that are given to the code and its users.
Besides, you also specifically say "We need to be able to commercialise it for our own costs", so you're not really "noncommercialisation", just anti-competitor.

> calling it "source available" means we're just opting into this mess of a debate around definitions

I'd say you're already opting yourself into more of a mess by using "open source" in a way that many consider this not to be. What mess are you opting into via using "source available"?

Alternatively, some use the term "fair code", see [the n8n case](https://github.com/n8n-io/n8n/issues/40) where they ended up using this term, after a very similar case to this.

> We would've chosen CC BY-NC [...] If you know a better license that addresses this situation (We need to be able to commercialise it for our own costs, but we dont want company to rip it off, slap a new label on it, put some features behind paywalls, and then sell it behind our backs, without any regard for our effort), then we'll might consider that one.

CC BY-NC is not open source either. The GPLv3 alone already does a good job of ensuring the code remains free to its users. A company could still "rip it off, slap a new label on it, put some features behind paywalls, and then sell it behind our backs" but they'd be required to attribute to you, and to provide sources to all they distribute to, including the added paywall features (Assuming those features are part of the same greater works). The AGPLv3 can add a little more protection in this area, by counting network use a distribution, but I'm not sure it at least immediately helps since this is not a hosted/SASS platform but a distributed app.

BrooklynMan:
GPL defines open source. Common clause clarifies that nobody can profit from our work.

If you cannot comprehend that none of the definition of open source means that you are entitled to profit from our work, then take your tantrum elsewhere. This matter is settled here.

PR #175, opened 2023-06-16
Centralized Settings with AppStorage support

tgeisse:
- added a centralized settings class so that the keys and defaults are not spread out across multiple files
- extended AppStorage to use the UserSettings keypath with support for all initializers. Benefit: multiple views can reference the same User Setting without having to know the UserDefaults key used nor the default value. The type (int, bool, enum, etc.) are also inferred so the user does not have to define that when declared
- also added a new property wrapper UserDefault. This provides the same interface for use in code outside of SwiftUI

# Checklist
- [X] I have described what this PR contains

- - [X] This PR does not introduce major changes
It provides a new interface for settings, but I do not see it as a major change to functionality.

- - [X] This PR does not change the UI in any way

# Pull Request Information

## About this Pull Request
AppStorage is great in SwiftUI for easy access to UserDefaults. However, it contains a flaw in easy reasuability across multiple views. Having to know the string key and default value for each use is less than ideal. 

This Pull Request provides a new UserSettings struct and Setting struct that allows for a central location for all UserDefaults to be defined. Now the key and default value are set in one location, making reuse easier.

AppStorage has also been extended so that KeyPaths can be used for AppStorage for use in SwiftUI views. The extensions are abstractions to the normal AppStorage initializers depending on the value of the default value of the setting. All AppStorage initializers have been implemented to support the KeyPath declaration method.

I've also provided a UserDefault method. While you can find simpler code online, those solution usually do not check that the types being used are compatable with UserDefaults. The implementation I've provided provides initializers similar to the AppStorage extensions. This has the added benefit that whatever settings you define will work in both @AppStorage and @UserDefault. UserDefault is intended to be used in non-SwiftUI code.

## Screenshots and Videos
Not UI impacting

## Additional Context
This PR is not necessary, but implementing something like this now will reduce headache before too many user settings are spread all over the place.


Mlem/Models/UserSettings.swift
```diff
@@ -0,0 +1,449 @@
...
+        
+        init(key: String) where Value: ExpressibleByNilLiteral {
+            self.key = key
+            self.defaultValue = nil
+        }
+    }
+}
+
+
+// MARK: - AppStorage Extension for Non-Optionals
```
ShadowJonathan, line 49: This is a lot of boilerplate that seems mostly the same pattern/code, is there a way to dedup that?

tgeisse: It does seem like a lot of boilerplate, but it is needed to match AppStorage's initializers: https://developer.apple.com/documentation/swiftui/appstorage

I can group extensions based on @available groupings and move where clauses, if that would be cleaner at least?

The UserDefault might be able to be reduced, but I submitted the PullRequest the way I did to match the initializers from AppStorage.

mormaer: I say we have a think about this, I like the idea - and I like the use of keypaths over string literals, but I wonder if we could achieve the same result with a wrapper?

Could we not have an object which declares our preferences, and accessible by keypath, like you've done here - but it just stores the values using `AppStorage` internally?

We wouldn't be able to declare a value that wasn't supported by `AppStorage` because the compiler would raise an error, avoiding the need to extend over every initialiser?

The only sticking point I can think of is the observability, but that should be solvable? 🤔

tgeisse: There is the solution that I have used before, but it is hard to get working with enums: https://www.avanderlee.com/swift/appstorage-explained/
The only way I have gotten it to work with enums (and nils) is to write explicit cases in the get / set accessors to check for my enums while casting.

It also does not provide type conformance, which means that trying to store a value that is not supported by UserDefaults will not fail until runtime.

The @UserDefault propretyWrapper I've written is not observable, true. The @AppStorage abstracted initializers are observable, though, in SwiftUI as the string counterparts are. This is where the alternative I've linked above comes at an advantage. I believe it can be used to observe changes outside of SwiftUI via Combine.

tgeisse: > Could we not have an object which declares our preferences, and accessible by keypath, like you've done here - but it just stores the values using AppStorage internally?
> 
> We wouldn't be able to declare a value that wasn't supported by AppStorage because the compiler would raise an error, avoiding the need to extend over every initialiser?

I just tried a bit and AppStorage does not expose, from what I can tell, anyway to observe changes.

tgeisse: I think I found a way. Please hold while I test it.

buresdv:
@ShadowJonathan & @mormaer, this PR is your call

mormaer:
I had a think about this and quickly hashed out a simpler approach that offers some of the same benefits, excuse the naming;

```swift
protocol Preference {
    associatedtype Value
    
    var key: String { get }
    var defaultValue: Value { get }
}

struct PreferenceThing<Value>: Preference {
    let key: String
    let defaultValue: Value
}

enum UserPreferences {
    static var shouldShowCompactPosts: PreferenceThing<Bool> { .init(key: "shouldShowCompactPosts", defaultValue: false) }
    static var voteComplexStyle: PreferenceThing<VoteComplexStyle> { .init(key: "voteComplexStyle", defaultValue: .standard) }
}
```

Usage:

```swift
    @AppStorage(UserPreferences.voteComplexStyle.key) var voteComplexStyle: VoteComplexStyle = UserPreferences.voteComplexStyle.defaultValue
```

---

I'm not a fan of the verbose-ness at the declaration, but we may be able to sprinkle some syntactic sugar on it to shorten it down?

I think this lets us define all our preferences in one place so a change to keys or default values becomes simple, it lacks the type safety in the respect that you _could_ send a value in that is not supported by `AppDefault`, we could perhaps constrain `Value` in the above to a new (empty) protocol like `AppDefaultCompatible`, and then extend the supported types with it to provide that safety...

I'd prefer something simpler like this personally, even if not quite as nice a keypath?

Mlem/MlemApp.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 @main
 struct MlemApp: App
 {
-    @AppStorage("hasUndergoneLegacyAccountDeletion_debug_3") var hasUndergoneLegaryAccountDeletion_debug: Bool = false
+    @AppStorage(\.hasUndergoneLegacyAccountDeletion_debug_3) var hasUndergoneLegaryAccountDeletion_debug
```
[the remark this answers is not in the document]
buresdv: Yes, it's no longer used and can be deleted

tgeisse:
I rewrote the implementation. I found a way to get the linked version working while also addressing the two major issues I had with it: lack of Enum support and lack of Optionals support. I was able to add both. Even better, it is done in a way that has no more boilerplate of long list of initializers. This is through the addition of a protocol that allows types to be extended and therefore easily added to the UserDefaults support.

Usage is easy:

In `UserDefault+Preference.swift` file, in the `Preferences` class, add a new UserDefaults backed item like this:
```swift
// Optional
@UserDefault("myNewOptional_UserDefaults_KeyName") var myNewOptional: String? // the '= nil' is optional

// Enum
@UserDefault("myEnum_UserDefaults_KeyName") var myEnum: MyEnum = .defaultCase

// Int
@UserDefault("myInt_UserDefaults_KeyName") var rateLimit = 500
```

To use the file in any file is simply:
```swift
@Preference(\.myNewOptional) var myOptionalString
```

SwiftUI views will rerender when the value changes. Non-SwiftUI can listen to changes this way:
```swift
class MyClass {
    @Preference(\.ud_Int) private var udInt
    private var cancellables: [AnyCancellable] = []
    
    init() {
        _udInt.publisher.sink { [weak self] _ in
            self?.performAction()
        }.store(in: &cancellables)
    }
    
    private func performAction() {
        // do something here when udInt changes
    }
}
```

ShadowJonathan:
Heya @tgeisse, we have a new repo at [mlemgroup/mlem](https://github.com/mlemgroup/mlem), due to some shuffling around we couldn't get it migrated properly, but if you could reopen this PR there, and link it back here, then it can continue to be considered ^^

tgeisse:
Sounds good!

PR #193, opened 2023-06-17
fix #179 - duplicate posts in feed

### LINKED ISSUE #179: Duplicate posts being added to feed
In some cases we are still ending up with duplicate posts in our `PostTracker` which is highlighted in Xcode by SwiftUI complaining that there are non-unique ids present in the view.

The posts themselves are identified by their id from the API and after discussion with @ShadowJonathan it's been agreed that these are uniquely identifying, this means the error is somewhere on our side.

There was a similar issue (https://github.com/buresdv/Mlem/issues/69) that has been resolved. In that case there were multiple calls being triggered for the same page, which has been resolved.

More investigation is needed to find where/how we're making a similar error.

mormaer:
# Checklist

# Pull Request Information

## About this Pull Request
While investigating #179 it was discovered that the API can return duplicate posts in subsequent calls to the `post/list` endpoint.

As this would require changes on the API side to resolve I have updated the `PostTracker` to reject duplicate posts being added to our list.

This will ensure the posts themselves remain uniquely identifiable which is a requirement when rendering in `SwiftUI`. We could have changed the `id` property on the `APIPostView` to be a `UUID` which would also meet this requirement, but this would then allow these duplicated posts to render and show which I believe is a bug in the API.

In addition to this I have moved responsibility for loading the feed into the `PostTracker` which removes the global `loadInfiniiteFeed(...)` function. The tracker will also then control it's `.isLoading` property to be `true` while a call is in progress.

Lastly, I've added a visual indicator that more posts are being loaded to the bottom of the feed while the tracker loads.

## Screenshots and Videos
| INDICATOR |
| --- |
| Screenshot 2023-06-18 at 15 50 47 |


Mlem/Models/Trackers/Post Tracker.swift
```diff
@@ -7,9 +7,75 @@
...
+    /// - Parameter newPost: The `APIPostView` you wish to add
+    func prepend(_ newPost: APIPostView) {
+        guard ids.insert(newPost.id).inserted else {
+            return
+        }
+        
+        posts.prepend(newPost)
+    }
+    
+    /// A method to supplly an updated post to the tracker
```
ShadowJonathan, line 64: Typo

ShadowJonathan (dismissed):
Very clean, LGTM if it works

Mlem/Models/Trackers/Post Tracker.swift
mormaer: doh, fixed 👍

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
```diff
@@ -465,26 +442,34 @@ struct CommunityView: View
...
+        if postTracker.isLoading {
+            VStack(alignment: .center) {
+                ProgressView()
+                    .frame(width: 16, height: 16)
+                Text("Loading more posts...")
+            }
+            .padding()
+            .frame(maxWidth: .infinity)
+            .foregroundColor(.secondary)
+            .background(Color.systemBackground)
```
WestonHanners, line 462: can you add
```
.accessibilityElement(children: .combine)
```

WestonHanners (changes requested):
Looks good, I like how clean it is now.

Please take a look at the one line I have requested for cleaning up accessibility for the new loader.

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
mormaer: added here -> fae8e187d30724b3f70e5a80b6c213c1ec1176cd

ShadowJonathan (approved):
LGTM

WestonHanners (approved):

PR #198, opened 2023-06-19
Guest mode and some API fixes

tht7:
- fixed a bug with the JSONDecoder in ApiClient that couldn't parse the response when the server sends inconsistent date formats (especially on devices with a different locale than the default en_US_POSIX)
- made the app a bit friendlier by adding a "Guest mode" now if you don't have any accounts you'll see a generic "All" feed, and even when you do have an account you can still enter "Guest mode" from the account list (we might want this in the future if we'd want to protect those accounts with FaceID and so a person handing his phone to his friend to check out the platform could remain private and enter "Guest mode")
- Made the account variable optional everywhere
- made sure to treat all the places that need the account (like upvoting) with a nice toast message

# Checklist

- - [x] This PR does not introduce major changes*

- - [x] This PR adds new UI elements / changes the UI, and I have attached pictures or videos of the new / changed UI elements (This PR hides certain buttons when you're in "Guest mode")

# Pull Request Information

## About this Pull Request
I've added a "Guest mode" to make the app friendlier, as well as fixed some bugs related to inconsistent data formatting from the server side

## Additional Context
While I've been a developer for many many years but I'm relatively new to Swift, I would love some feedback, I would love to help and contribute to this app even if it means I'll have to learn a bit before ;)

### DIFF
--- a/Mlem/API/APIClient.swift
+++ b/Mlem/API/APIClient.swift
@@ -18,6 +18,8 @@ enum APIClientError: Error {
     case response(APIErrorResponse, Int?)
 }
 
+let DefaultLemmyServer = URL(string: "https://lemmy.ml/api/v3/")!
+
 class APIClient {
     
     let session: URLSession
@@ -40,6 +42,12 @@ class APIClient {
             let statusCode = (response as? HTTPURLResponse)?.statusCode
             throw APIClientError.response(apiError, statusCode)
         }
+        
+        decoder.dateDecodingStrategyFormatters = [ DateFormatter.expandedTWithTimezone,
+                                                   DateFormatter.expandedT,
+                                                   DateFormatter.standardT,
+                                                   DateFormatter.standard,
+                                                   DateFormatter.yearMonthDay ]
         
         return try decoder.decode(Request.Response.self, from: data)
     }
--- a/Mlem/API/Requests/SearchRequest.swift
+++ b/Mlem/API/Requests/SearchRequest.swift
@@ -26,7 +26,7 @@ struct SearchRequest: APIGetRequest {
     let queryItems: [URLQueryItem]
 
     init(
-        account: SavedAccount,
+        account: SavedAccount?,
         query: String,
         searchType: SearchType,
         sortOption: SortingOptions = .topAll,
@@ -37,9 +37,9 @@ struct SearchRequest: APIGetRequest {
         creatorId: Int? = nil,
         limit: Int? = nil
     ) {
-        self.instanceURL = account.instanceLink
-        var queryItems: [URLQueryItem] = [
-            .init(name: "auth", value: account.accessToken),
+        self.instanceURL = account?.instanceLink ?? DefaultLemmyServer
+        let queryItems: [URLQueryItem] = [
+            .init(name: "auth", value: account?.accessToken),
             .init(name: "type_", value: searchType.rawValue),
             .init(name: "sort", value: sortOption.rawValue),
             .init(name: "listing_type", value: listingType.rawValue),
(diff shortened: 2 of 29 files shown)


mormaer:
Hey @tht7 👋 

Would you be able to raise a separate PR with just the date decoding changes included?

I'd like to get that in as soon as possible if you have the time 🤞 

Guest mode was raised recently and I got the impression that it wasn't to be included. Personally, I think it would be a positive addition, but there are likely some UX discussions to be had around it before decisions are made... I'll leave that open to others to weigh in on, what are your thoughts on guest mode @buresdv?

Mlem/API/Requests/SearchRequest.swift
```diff
@@ -37,9 +37,9 @@ struct SearchRequest: APIGetRequest {
         creatorId: Int? = nil,
         limit: Int? = nil
     ) {
-        self.instanceURL = account.instanceLink
-        var queryItems: [URLQueryItem] = [
-            .init(name: "auth", value: account.accessToken),
+        self.instanceURL = account?.instanceLink ?? DefaultLemmyServer
+        let queryItems: [URLQueryItem] = [
+            .init(name: "auth", value: account?.accessToken),
```
ShadowJonathan, line 42: Don't make null-tolerant versions of these APIs, make additional .init's which take only the Instance URL, so that the client a layer up can provide it that default URL if it wishes so

Else this will most likely just lead to buggy and inconsistent behaviour

Mlem/API/APIClient.swift
```diff
@@ -41,6 +43,12 @@ class APIClient {
             throw APIClientError.response(apiError, statusCode)
         }
         
+        decoder.dateDecodingStrategyFormatters = [ DateFormatter.expandedTWithTimezone,
+                                                   DateFormatter.expandedT,
+                                                   DateFormatter.standardT,
+                                                   DateFormatter.standard,
+                                                   DateFormatter.yearMonthDay ]
+
```
ShadowJonathan, line 51: I don't know if this PR will be merged quickly, but this might be useful on its own, could you separate this in its own PR?

ShadowJonathan (changes requested):
We don't want logic like this on the API layer, we want that as declarative and clear as possible.

Also, like @mormaer said, we recently have gotten the request for guest mode as well, but at this stage, we're not so sure

Mlem/API/APIClient.swift
tht7: I have! https://github.com/buresdv/Mlem/pull/208

Mlem/API/Requests/SearchRequest.swift
tht7: I wanted to leave the UI layer blissfuly unaware when the account is nil and having the API layer have a sensible fallback seemed like the way to go, 
I'm not really sure how to implement this cleanly in the UI layer without adding this parameter, and passing it down (with a fallback in case the account is nil) on every request, which to me looks dirty, but if you think it's the way I would be happy to change it 😄

JakeShirley:
I can see a world where there is a "Discover a server" experience in the app.  Lemmy is definitely less than user friendly.

buresdv:
I can see a "guest mode" working is if we choose a specific set of instances (further discussion is warranted to choose which instances they will be), give them some context (for example "this is a homepage for tech nerds", "this is a homepage for general discussion", etc.), and then let the user pick from them. We should make it discoverable to the average user without them having to scour for instances.

Before a guest mode is implemented, we also have to implement a registration workflow so the users can sign up to their chosen instance right from the app.

ShadowJonathan:
Random; We could probably also hook this into the signup flow, i.e. now have a user scour for instances but allow them to pick from a list, which new instance to sign up at.

mormaer:
I've created an issue for discussion to continue on this here -> #216

I think for now we may not be able to proceed with this PR until we spend some time ironing out what discovery/guest mode would look like

mormaer:
@tht7 - will we close this PR for now until we decide on the direction we want to go in for guest mode / discovery mode?

PR #207, opened 2023-06-20
Updated user role flair

JakeShirley:
# Checklist

- - [ ] This PR does not introduce major changes
- - [x] This PR introduces major changes, and I have consulted @buresdv, @jboi or @mormaer in the *Mlem Development Matrix room*

# Pull Request Information

## About this Pull Request
This PR updates the User Link View to take a post and comment context so they can determine how to render the user.  They are both optional.

You can see the flair types in the image bellow.

The developers list is just a hardcoded list in the view which we can update with new people as they contribute.

This PR also adds a ton of dev-crafted data at the bottom to maintain the preview.  We should consider a long term home for these as they are nice utilities to have and would make previewing the other views more possible.

## Screenshots and Videos
From the preview:
image

In action:
image

## Additional Context
Updated `APIPerson` to haven an equality operator


Mlem/API/Models/Users/APIPerson.swift
```diff
@@ -29,3 +29,9 @@ struct APIPerson: Decodable {
     let banExpires: Date?
     let instanceId: Int
 }
+
+extension APIPerson: Equatable {
+    static func == (lhs: APIPerson, rhs: APIPerson) -> Bool {
+        lhs.actorId == rhs.actorId
```
ShadowJonathan, line 35: This looks ok, but why was it necessary? .id works fine too, since every person has an unique IDs in the server we're talking with

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -14,7 +14,24 @@ struct UserProfileLink: View
...
+    
+    // Extra context about where the link is being displayed
+    // to pick the correct flair
+    @State var postContext: APIPostView? = nil
+    @State var commentContext: APIComment? = nil
+    
+    static let developerNames = [
+        "lemmy.ml/u/lFenix",
+        "vlemmy.net/u/darknavi",
+        "lemmy.ml/u/BrooklynMan"
```
ShadowJonathan, line 26: beehaw.org/u/jojo for me ^^

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -26,24 +43,117 @@ struct UserProfileLink: View
...
-                    .foregroundColor(.yellow)
+        }
+        if user.admin {
+            return UserProfileLink.flairAdmin
+        }
+        if user.botAccount {
+            return UserProfileLink.flairBot
+        }
+        if commentContext != nil {
+            if commentContext!.distinguished {
```
ShadowJonathan, line 85: you can do `if let commentContext` and it essentially does nil check and the unwrapping, safely

ShadowJonathan (commented):
Looking good, a few comments

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -26,24 +43,117 @@ struct UserProfileLink: View
...
+        if postContext != nil {
+            if user == postContext!.creator {
+                return UserProfileLink.flairOP
+            }
+        }
+        return nil
+    }
+}
+
+// TODO: darknavi - Move these to a common area for reuse
```
mormaer, line 98: Apple did introduced a common area for this, and they're then stripped from the bundle for production too 🙌

https://www.avanderlee.com/xcode/development-assets-preview-catalog/

It's worth a discussion amongst the team as to how we approach providing mock data for more complex previews, we might want to consider having blobs of JSON in the development assets for example, and then similar to the linked article we could extend the types we care about with a `.preview` property or such to pass in to the preview area?

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -14,7 +14,24 @@ struct UserProfileLink: View
...
+    static let developerNames = [
+        "lemmy.ml/u/lFenix",
+        "vlemmy.net/u/darknavi",
+        "lemmy.ml/u/BrooklynMan"
+    ]
+    
+    static let flairDeveloper = UserProfileLinkFlair(color: Color.purple, systemIcon: "hammer.fill")
+    static let flairMod = UserProfileLinkFlair(color: Color.green, systemIcon: "shield.fill")
+    static let flairBot = UserProfileLinkFlair(color: Color.indigo, systemIcon: "server.rack")
+    static let flairOP = UserProfileLinkFlair(color: Color(red: 249 / 255, green: 105.0 / 255, blue: 14.0 / 255), systemIcon: "person.fill")
```
mormaer, line 32: we should aim to avoid defining colours like this in the code as it can become a headache for light/dark mode down the line.

If we can't use a system defined colour, we should agree some custom colours (for both light/dark) with design and add then to our asset catalogs, that way we can reference them by name throughout the project and they will automatically adjust based on the users interface setting 🙏

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -26,24 +43,117 @@ struct UserProfileLink: View
...
-                    .foregroundColor(.indigo)
+        }
+    }
+    
+    struct UserProfileLinkFlair {
+        var color: Color
+        var systemIcon: String? = nil
+    }
+    
+    private func calculateFlair() -> UserProfileLinkFlair? {
```
mormaer, line 72: it could be worth making the `UserProfileLinkFlair` a standalone struct with an initialiser that takes the parameters it requires, the initialiser can then do much the same logic as `calculateFlair` and return the correct value?

if it was stand-alone it'd make re-using it elsewhere easier too 🙏

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
JakeShirley: JSON blobs sounds good to me, although some code-generated stuff would be useful to quickly generating lots of cases (different types of posts, etc.)

I'll poke around the dev assets catalog when I get a chance!

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
JakeShirley: I'll add them to the assets catalog, good call!

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
JakeShirley: I just tweaked the name, but this method actually calculates both the color AND the icon ("flair") so I'm not sure if spinning this off to its own view makes much sense.  The `UserProfileLink` its self is a pretty stand-alone, reused view though.

ShadowJonathan (approved):
LGTM, given a UX pass

mormaer:
Raising this not as a blocker, but as something I think we should address ASAP for the project going forward...

Apologies, I was unclear with my comment about the colours - I was specifically meaning the colours that had been explicitly defined, but the recent changes has moved them all the live in the asset catalog, but they're not adaptive... it's the same colour defined for both themes.

Colours provided by Apple such as `.green`/`.red`/`.purple` are adaptive out of the box, so are designed to look good in the different modes. With these changes we're using the same colours for both, they don't look _bad_ but the contrast is impacted.

| THEME | BEFORE | AFTER |
| --- | --- | --- |
| Light | Screenshot 2023-06-20 at 17 18 47 | Screenshot 2023-06-20 at 17 19 21 |
| Dark | Screenshot 2023-06-20 at 17 18 31 | Screenshot 2023-06-20 at 17 19 38 |

I think we should aim to use Apple's colours if possible, and if we want to introduce our own then we should make them adaptive by defining different tones which fit the theme.

Are we able to consult with https://github.com/BrooklynMan and @WestonHanners regarding custom colours that are adaptive and accessible?

BrooklynMan:
i agree with this completely. official ios colors should be used.

mormaer:
@BrooklynMan - are you able to advise which colours you'd like to see used for the different flairs from the list Apple show here:

https://developer.apple.com/design/human-interface-guidelines/color#iOS-iPadOS-system-colors

if none of those fit, we could add custom ones if we have a light/dark variant defined 🙏

Mlem/Views/Tabs/Posts/Components/Shared/Links/User Profile Link.swift
```diff
@@ -27,11 +27,11 @@ struct UserProfileLink: View
         "beehaw.org/u/jojo",
     ]
     
-    static let flairDeveloper = UserProfileLinkFlair(color: Color("UserFlairDeveloper"), systemIcon: "hammer.fill")
-    static let flairMod = UserProfileLinkFlair(color: Color("UserFlairModerator"), systemIcon: "shield.fill")
-    static let flairBot = UserProfileLinkFlair(color: Color("UserFlairBot"), systemIcon: "server.rack")
+    static let flairDeveloper = UserProfileLinkFlair(color: Color.purple, systemIcon: "hammer.fill")
+    static let flairMod = UserProfileLinkFlair(color: Color.green, systemIcon: "shield.fill")
+    static let flairBot = UserProfileLinkFlair(color: Color.indigo, systemIcon: "server.rack")
     static let flairOP = UserProfileLinkFlair(color: Color("UserFlairOP"), systemIcon: "person.fill")
```
ShadowJonathan, line 33: Could this color also be apple-fied?

JakeShirley:
Here is what we landed on today

image
image

ShadowJonathan (approved):
LGTM

WestonHanners:
Agreed. Using apple colors side steps some contrast concerns. It also automatically adjusts for increased contrast preferences.

mormaer (approved):
I'm personally not a fan of calling out Mlem developers with a flair as it could he confusing for a _general user_ and give additional weight to their comments, but that's just me and a personal thing 🙃 

I'd like to see it as a feature of our community via Lemmy directly instead rather than flair across the whole of Lemmy, but that's a future thing and doesn't impact this PR for me 😄

PR #208, opened 2023-06-20
fixed a bug in the JSONDecoder in the ApiClient

tht7:
- fixed a bug in the JSONDecoder in the ApiClient that couldn't parse the response when the server sends inconsistent date formats (especially on devices with a different locale than the default en_US_POSIX)

# Checklist

- - [x] This PR does not change the UI in any way

# Pull Request Information

## About this Pull Request

## Screenshots and Videos

## Additional Context
Hmmm I messed up the formatting and my second commit should've fixed that but it still looks like a bit of a mess, sorry, where was actually very little code changed


Mlem.xcodeproj/project.pbxproj
```diff
@@ -1477,6 +1484,7 @@
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				ENABLE_NS_ASSERTIONS = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				ENABLE_USER_SCRIPT_SANDBOXING = YES;
```
ShadowJonathan, line 1487: There are additional lines like this, plus the version check

I don't know if they'll interfere badly with other people's setups, and this sounds like a security risk, could you remove them?

Mlem/API/APIClient.swift
```diff
@@ -18,6 +18,8 @@ enum APIClientError: Error {
     case response(APIErrorResponse, Int?)
 }
 
+let DefaultLemmyServer = URL(string: "https://lemmy.ml/api/v3/")!
```
ShadowJonathan, line 21: Leftover

ShadowJonathan (commented):
Looks good, except for a few nits

Mlem.xcodeproj/project.pbxproj
tht7: Removed

Mlem/API/APIClient.swift
tht7: Removed! sorry about it

ShadowJonathan (dismissed):
LGTM, though I want to have mormaer (the original writer of this code) take another look at this

mormaer:
if we can hold for now @ShadowJonathan, I do have a few suggestions and also these changes as they are would re-introduce the offset bug as their timezone is not set to `.gmt` like the current one.

I'm having quite a busy day but should have time to review properly this afternoon 🤞

ShadowJonathan:
No worries, thanks for spotting, this is why I requested additional review :)

Mlem/Extensions/JSONDecoder+dateDecodingStrategyFormatters.swift
```diff
@@ -0,0 +1,82 @@
...
+        var dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+        dateFormatter.timeZone = TimeZone(identifier:"GMT")
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        return dateFormatter
+    }()
+
+    static let yearMonthDay: DateFormatter = {
+        var dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd"
```
mormaer, line 44: which Lemmy instance have you seen returning this format from the API?

it'd be worth us having a list of any instances we come across who have these  _rogue_ responses so that we can ensure we test builds against them to make sure we're as compatible as possible

mormaer (changes requested):
I like the approach of being able to pass a set of date formatters in via extension, however in our current use case (and hopefully forever!) the only place we're expecting to need to parse a `String` to a `Date` is as part of the decoding process from the API response...

We already have an extension on `JSONDecoder` in the file `JSONDecoder+Default` which is the one the client is using... rather than declare multiple different date formatters given the only thing that will change is the format string we're going to use, would it be a simpler change to adapt our existing extension to include your locale fix and then loop over any formats we declare - it'd look like the below:

```swift
            // ... existing implementation ...

            formatter.timeZone = .gmt
            formatter.locale = Locale(identifier: "en_US_POSIX")
            
            let formats = [
                "yyyy-MM-dd'T'HH:mm:ss.SSSSSS",
                "yyyy-MM-dd'T'HH:mm:ss",
                // other required formats we've seen from Lemmy APIs
            ]
            
            for format in formats {
                formatter.dateFormat = format
                if let date = formatter.date(from: string) {
                    return date
                }
            }

           // ... existing implementation ...
```

I think this could be slightly cleaner given it's only the format string we want to vary while decoding?

Mlem/Extensions/JSONDecoder+dateDecodingStrategyFormatters.swift
tht7: I didn't keep a list but I'll start doing so

tht7:
@mormaer awesome feedback! I've implemented all your suggestions, 

I feel a bit silly for missing the already existing extension,
 but my hope is that with time (and maybe a few more contributions) I will get more familiar with the code base.

So thank you for the patience

mormaer (approved):
Looks good to me 🎉

mormaer:
> I feel a bit silly for missing the already existing extension, but my hope is that with time (and maybe a few more contributions) I will get more familiar with the code base.

The structure is a bit messy so I wouldn't worry 🙃 

It's a great first contribution and fixes a major bug - so thanks 🎉

Mlem/Extensions/JSONDecoder+Default.swift
```diff
@@ -11,35 +11,33 @@ extension JSONDecoder {
...
+        
+        let formats = [
+            "yyyy-MM-dd'T'HH:mm:ss.SSSSSS",
+            "yyyy-MM-dd'T'HH:mm:ss.SSSSZ",
+            "yyyy-MM-dd'T'HH:mm:ss.SSSS",
+            "yyyy-MM-dd'T'HH:mm:ss",
+            "yyyy-MM-dd HH:mm:ss.SSSSSS",
+            "yyyy-MM-dd HH:mm:ss",
+            "yyyy-MM-dd"
+        ]
```
ShadowJonathan, line 28: FWIW there's only a few kinds of date formats that Lemmy is expected to return, but this should work

Separately; if it hits the last one, is it possible it'll only parse the first part of the string, and disregard the rest? Or should it match exactly?

ShadowJonathan (approved):
LGTM, one last comment

Mlem/Extensions/JSONDecoder+Default.swift
tht7: I think It should match exactly unfortunately, which is why I added it as a last resort

mormaer: yeah, the formatter is very specific so the last would only map to dates in that exact format, I agree that I don't think we've ever seen that (or should see it), with it being last in the list it likely won't hit it often, other than before failing.

Happy to get this merged if you are @ShadowJonathan?

ShadowJonathan: If it maps in an exact format, it should be fine, approved.

WestonHanners (approved):

PR #214, opened 2023-06-21
Faster image loading

tht7:
# Checklist

# Pull Request Information

## About this Pull Request
- made the feed pre-load a lot more posts (50 which seems to be the limit on almost all lemmy servers)
- started loading the next page of a feed 25 posts before the user reaches the end of the feed
- created a new app-wide cache of up to 500MiB (in memory and on disk)
- when loading new posts added a function to pre=load all images into cache in the background (so the user doesn't experience the annoying image loading jitters)
- added a setting in the app to see how much space the cache is taking and allowing the user to clear it

All these changes made scrolling in the app a lot smoother! (but from my testing I think we should change point number 2 to start loading new posts 49 posts in advance, from my testing we start out by loading a few pages and it gives the user a much much smoother experience)

## Screenshots and Videos
Added a new button in the settings
image

## Additional Context


mormaer:
Thanks for another great contribution 👍 

I'll try and get a review done on this later - definitely something we should be looking to integrate, my only immediate hesitation is bumping the feed limit up as high as `50`, at the moment I think we usually get ~10, so a smaller bump (plus your image pre-loading) would likely give much better scrolling UX without as much strain on slower connections?

Emiliaaah:
Something to consider is to load a smaller amount of posts on the first load and (pre-)load a larger amount after that. This is the way Apollo does it. I think it's a pretty smart way of doing this, since you get the first few posts quickly which makes everything more responsive. You also get the benefit of not having to send as many API requests. Then you can also decide if you want to immediately pre-load the next set of posts, so it has a bit of time to fetch those on a slower connection.

tht7:
I completely understand your concerns

But I have a few counterpoints (and hopefully some receipts to back them up 😉)

You’re correct that under stricter network conditions loading will be slightly longer but the time per post is significantly greater which leads to a way nicer experience in the end
Plus, I start loading the next batch of posts way before you reach the actual end of the feed giving the network enough time to catch up

Here are a few measurements that I’ve captured:

Very good network conditions:
10 posts loaded in 1.350360666 seconds (0.1350360666s per post)
50 posts loaded in 0.538415459 seconds (0.01076830918s per post)
50 posts loaded in 0.63201625 seconds (0.012640325s per post)
50 posts loaded in 0.466469417 seconds (0.00932938834s per post)
50 posts loaded in 0.411641667 seconds (0.00823283334s per post)
50 posts loaded in 1.064755458 seconds (0.02129510916s per post)
50 posts loaded in 1.038863958 seconds (0.02077727916s per post)
50 posts loaded in 0.482592583 seconds (0.00965185166s per post)
(in this capture I wanted to illustrate that under reasonable network conditions the batch size makes almost no difference in the loading time, but it does make a huge user experience difference because we simply need to load a lot less batches)

3G:
10 posts loaded in 1.925274167 seconds (0.1925274167s per post)
50 posts loaded in 1.76143075 seconds (0.035228615s per post)
50 posts loaded in 2.635654958 seconds (0.05271309916s per post)
50 posts loaded in 5.396175124999999 seconds (0.1079235025s per post)
50 posts loaded in 1.694496958 seconds (0.03388993916s per post)
50 posts loaded in 2.154538083 seconds (0.04309076166s per post)
50 posts loaded in 5.186091708999999 seconds (0.1037218342s per post)
50 posts loaded in 5.500505959000001 seconds (0.1100101192s per post)
50 posts loaded in 4.257322167 seconds (0.08514644334s per post)
(here we do see that loading bigger batches usually take longer but not per post so when the loading is done you have more to scroll through before you need to load again, and because the bash size here is the largest we can make it, I also start preloading a lot sooner which leads to silky, smooth scrolling, even under kind of terrible network conditions)

Personal note from my personal experience:
In my profession, I am a backend engineer, and I have been a backend engineer for about five years now (maybe even more, but who counts)

I have grown this approach of ”data generosity” and whenever I interact with the frontend engineers in my company, I encourage them to ask larger batches if they know they’re gonna use this data

Every HTTP request has its overheads,
TSL, SSL, authentication, pinging the database, and if the service is written in another language, there’s also parsing and formatting the responses

By asking for larger batches of data the server needs to perform a lot less of these menial tasks and we can focus on what’s really important: the data itself!

I urge you to clone this branch and give it a try because it’s a transformative experience how smooth and nice the feed scrolls the larger the data batches get,

It is so nice that I consider Mlem from TestFlight without this patch, barely usable, and when I run Mlem for personal consumption, I crank all the numbers up to 50 all of the time to preload as much as possible 

Of course, I leave the decision in your hands, and I’ve even experimented with making the first batch smaller (back to the original 10 posts) only to find it doesn’t make any significant difference, except it does slow things down a bit

Mlem/App Constants.swift
```diff
@@ -11,22 +11,24 @@ import UIKit
 
 struct AppConstants
 {
+    static let cacheSize = 500_000_000 // 500MiB in bytes
+    static let urlCache: URLCache = URLCache(memoryCapacity: cacheSize, diskCapacity: cacheSize)
```
WestonHanners, line 15: 500MB memory for images? seems a tad excessive.

tht7: You might have a point 🤔 the maximum cache size of Apollo is 200 MB

And after scrolling my feed until my finger hurt, I only got to 148 MB of cache usage 

It's important to note though, that this cache is managed by iOS and will be cleared if the device is low on memory automatically (it will be cleared from disk If the device is running out of storage, or if the application is requesting too much memory, it’s also a reason for iOS to clear it from memory)

Plus, we have the button in the settings to clear it if the user wants to make more space for other stuff 

To me those two points justify having a larger cache (since it will be cleared without hesitation if it bothers the user at any point)

But I leave this decision in your capable hands 😉

BrooklynMan: Apollo kept that cache size small with a mix of reddit doing a ton of heavy lifting with all manner of image optimization for almost every imaginable use case, and Apollo employing pretty aggressive cache and active memory management. images in your feed, for example, got flushed pretty quickly when you were scrolling, and only ones recently in your viewport remained when you tapped on a post. if you were gone from your feed long enough, they got dumped. even comments in a post didn't hang around for long and, when swiping pack to your feed and scrolling again, a lot would have to be reloaded.

I would run into this problem a lot when I would ride the subway and have no connection between stations.

WestonHanners (approved):

Mlem/Models/Trackers/Post Tracker.swift
```diff
@@ -25,53 +25,83 @@ class PostTracker: ObservableObject {
...
-        
+
         let request = GetPostsRequest(
             account: account,
             communityId: communityId,
             page: page,
             sort: sort,
-            type: type
+            type: type,
+            limit: 50
```
mormaer, line 35: based on the comment about trying to make the initial load as fast as possible we could supply a slighter lower value for the first page by doing:
```swift
limit: page == 1 ? 25 : 50
```
It may only save a small fraction of a second on that first load, but the faster the user sees that initial content the better, right?

Mlem/Models/Trackers/Post Tracker.swift
```diff
@@ -25,53 +25,83 @@ class PostTracker: ObservableObject {
...
+                preloadSingleImage(url: url)
+            default:
+                break
+            }
+
+        }
+    }
     
+    private func preloadSingleImage(url: URL) {
+        let task = URLSession.shared.dataTask(with: url) { data, res, err in
```
mormaer, line 68: can we convert this to use the `async` variation instead of the completion blocks? We're using `async` elsewhere throughout the project:

```swift
private func preloadSingleImage(url: URL) async {
    do {
        let request = URLRequest(url: url)
        let (data, response) = try await URLSession.shared.data(for: request)
        let cachedResponse = CachedURLResponse(response: response, data: data)
        AppConstants.urlCache.storeCachedResponse(cachedResponse, for: request)
    } catch {
        // no action required on failure...
    }
}
```
we'll need to add a few `await`'s further up and mark the `preloadImages(...)` as `async` but we should try and stick with the newer APIs from Apple 🙏

Mlem/Models/Trackers/Post Tracker.swift
```diff
@@ -25,53 +25,83 @@ class PostTracker: ObservableObject {
...
+        Task(priority: .background) {
+            preloadImages(response.posts)
+        }
+
         add(response.posts)
         page += 1
     }
+
+    func preloadImages(_ newPosts: [APIPostView]) {
+        URLSession.shared.configuration.urlCache = AppConstants.urlCache
```
mormaer, line 52: This doesn't need changed in this PR, but once this is in it'd be an idea to move the pre-loading responsibility into a separate object, to me it feels we're muddying the responsibility of the post tracker somewhat with this, I think we should aim to keep the tracker solely focussed on handling posts, and have something else deal with image and other types of caching behaviours going forward

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
```diff
@@ -462,13 +462,17 @@ struct CommunityView: View
             }
             .buttonStyle(EmptyButtonStyle()) // Make it so that the link doesn't mess with the styling
             .task {
-                if post == postTracker.posts.last {
-                    await loadFeed()
+                if !postTracker.isLoading {
+                    if let position = postTracker.posts.lastIndex(of: post) {
+                        if  position >= (postTracker.posts.count - 40) {
```
mormaer, line 467: the threshold here could likely be a bit closer to the bottom of the feed, but again I don't think it needs to change for this PR.

The API supports the ability to [mark posts as read](https://join-lemmy.org/api/interfaces/MarkPostAsRead.html), so we may want to look at having the post tracker be informed when posts are viewed to support that feature, at which point it could have the responsibility for deciding when the load the next page to remove this logic from the view all together.

Again nothing to change here, I'm just thinking out loud 🙃

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
```diff
@@ -514,7 +518,8 @@ struct CommunityView: View
         } catch {
             // TODO: we may be receiving decoding errors (or something else) based on reports in the dev chat
             // for now we will fail silently if the user has posts to view while we investigate further
-            assertionFailure("Unhandled error encountered, if you can reproduce this please raise a ticket/discuss in the dev chat")
+            print(String(describing: error))
+//            assertionFailure("Unhandled error encountered, if you can reproduce this please raise a ticket/discuss in the dev chat")
```
mormaer, line 522: can we put this assertion back in instead of the `print` please? 🙏

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Post Item/Components/Compact Post.swift
mormaer, line 1: should this be in the changes?

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Post Item/Components/Large Post.swift
mormaer, line 1: same question here, both files show as empty?

Mlem/Views/Tabs/Settings/Components/Views/General.swift
```diff
@@ -92,6 +93,30 @@ struct GeneralSettingsView: View
...
 
             }
+
+            Section()
+            {
+                Button(role: .destructive) {
+                    URLCache.shared.removeAllCachedResponses()
+                    diskUsage = Int64(URLCache.shared.currentDiskUsage)
+                } label: {
+                    Label("Image cache: \(ByteCountFormatter.string(fromByteCount: diskUsage, countStyle: .file))", systemImage: "trash")
```
mormaer, line 103: should we just refer to this as `Cache`, although we're explicitly using it for images we're using the systems `.shared` instance, so it will be storing more than images potentially

Mlem/Views/Tabs/Settings/Contributors/Contributors View.swift
```diff
@@ -17,7 +17,7 @@ struct ContributorsView: View
     {
         VStack(alignment: .center, spacing: 10)
         {
-            CachedAsyncImage(url: contributor.avatarLink)
+            CachedAsyncImage(url: contributor.avatarLink, urlCache: AppConstants.urlCache)
```
mormaer, line 20: Nothing to change for this PR here, but we may want to build a small wrapper around `CachedAsyncImage` which automatically sets our cache to save having to remember to do it everywhere going forward?

mormaer (changes requested):
Sorry it's taken so long to get round to reviewing this - some small things to change if you have time 🙏 

Looking forward to seeing this go in 🎉

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
tht7: done!

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Post Item/Components/Compact Post.swift
tht7: dammit, it's weird WebStorm formatting, I found the GIT tools in Xcode a bit weird so I swapped to my lovely Webstorm, until now I always did "selective commits" and carefully deselected those formatting differences but I guess it got  the better of me this time 😓

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Post Item/Components/Large Post.swift
tht7: same here

Mlem/Views/Tabs/Settings/Components/Views/General.swift
tht7: Yea but the idea was that maybe we'd want to have more then one cache (think image cache and videos cache so a user can clear the massive video cache but not have to suffer all the images reloading as well)

but for now I've updated it

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Post Item/Components/Compact Post.swift
tht7: Ok I think I've manager to revert that

Mlem/Views/Tabs/Settings/Components/Views/General.swift
mormaer: > Yea but the idea was that maybe we'd want to have more then one cache

Yeah, my thinking too - but while we have one it's best to keep the wording generic 👍

mormaer (approved):
Looks great! 🎉

PR #229, opened 2023-06-22
common clause license file added

BrooklynMan:
I have also added the Common Clause 1.0 license text file, which includes our org info and app name.


WestonHanners (approved):

lil5:
This is already added in the LICENSE file. This pr is unnecessary

PR #233, opened 2023-06-23
Added CONTRIBUTION.txt, EmbeddedPost

EricBAndrews:
Added CONTRIBUTING.txt, tweaked swipiness

### DIFF
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -1,0 +1,19 @@
+Hello!
+
+If you're reading this, you probably want to contribute to Mlem. Welcome! We're happy to have you on board.
+
+First and foremost, you should join our Matrix room: https://matrix.to/#/#mlemapp-recruitment:matrix.org. The people there can give you a good idea of what work needs doing.
+
+To avoid having multiple in-flight tasks working on the same part of the codebase, we have a set procedure for claiming and performing work. If you don't follow it, your PR will *probably* be rejected (unless it's really *that* good).
+
+1. Go to our project board and find something in the "todo" section
+2. Comment that you want to handle that task
+3. !!! WAIT !!! for the task to be assigned to you! This step is *very* important!
+4. Fork the repository and develop the change on your fork
+5. Open a PR for the task. The description should clearly reference the issue that you are addressing, and your PR should be able to merge with no conflicts--if the master branch changes before your PR is merged, you should either rebase or merge so that your fork can be merged cleanly.
+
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.txt
```diff
@@ -0,0 +1,19 @@
...
+To avoid having multiple in-flight tasks working on the same part of the codebase, we have a set procedure for claiming and performing work. If you don't follow it, your PR will *probably* be rejected (unless it's really *that* good).
+
+1. Go to our project board and find something in the "todo" section
+2. Comment that you want to handle that task
+3. !!! WAIT !!! for the task to be assigned to you! This step is *very* important!
+4. Fork the repository and develop the change on your fork
+5. Open a PR for the task. The description should clearly reference the issue that you are addressing, and your PR should be able to merge with no conflicts--if the master branch changes before your PR is merged, you should either rebase or merge so that your fork can be merged cleanly.
+
+In addition, please develop according to the following principles:
+- One named View struct per file. The name of the file should describe the view (e.g., "Large Post,"), and the name of the struct should be match the name of the file but with spaces removed (e.g., "LargePost"). Avoid ending these file and struct names with "View"; that is implicit in their location under the Views/ folder
```
mormaer, line 16: My two cents;

Personally, I think files should be named to match the class/struct/thing they contain so if you created a view called `ExampleView` it would live in a file call `ExampleView.swift`

Also, as above the struct would have view on the end, it may be implicit because it lives under a folder called Views in the project, but you don't always have that context when it appears elsewhere in the codebase, say for example you were creating an extension, or using it as an argument, it's type would appear as just `Example` and without navigating to it to check what it is it's ambiguous 🤷‍♂️

WestonHanners: I agree, especially once we are looking at CI/CD logs, Settings (the view) and Settings (the model) might get confusing.

EricBAndrews: Hmm, OK, that make sense to me. I'll adjust it to that views should always have "view" in the name and open an issue to enforce that naming scheme. Our *entire* codebase has file names with spaces, so unless we have strong feelings there I'm keeping the "spaces in the title" stipulation.

WestonHanners (approved):

CONTRIBUTING.txt
mormaer: I have _very_ strong feelings about the spaces in our file names, I'm not sure what triggers me more, the spaces or the `{`'s on their own lines 😂

Issue #260, opened 2023-06-24
extra security for the app


AliMickey:
## Introduction
Other reddit clients had the support for either faceid/touchid + plus optional custom passcode to open the app. 

## Requirements
An option to enable extra security.

PR #316, opened 2023-06-26
FaceID TouchID and friends

tht7:
# Checklist

- - [ ] This PR does not introduce major changes
- - [x] This PR introduces major changes, and I have consulted the *Mlem Development Matrix room*

# Pull Request Information
Wop wop made a new place to place all changing account preferences that's way more reactive with SwfitUI
added a new option to lock an account with FaceID

## Screenshots and Videos

https://github.com/mormaer/Mlem/assets/6067331/9c9c665f-f7a6-48d9-a6a2-7b76a7827bfe

## Additional Context
Maybe #260


ShadowJonathan (changes requested):
The blur is clear enough that someone can still somewhat make out the things behind it, maybe make it even more blurred?

tht7:
I found your comment kinda funny buttt I loved it so, there you go it’s the same screen (of the Mlem community) but now with iOs’s thiccMaterial image

ShadowJonathan (dismissed):
Thanks, no further nits, i'll leave @JakeShirley to review this further :)

mormaer:
should a blocking screen like this not be presented as a `.fullScreenCover`?

in the video we're only covering the central content, the navigation and tab bars remain visible which show potentially sensitive information like the users name and community names etc?

Is interaction blocked on those visible elements?

mormaer:
Another small thought, generally biometric locks like this on iOS will attempt authentication automatically when you open, as opposed to requiring the user to interact by tapping?

tht7:
I was still working on this, but in order to achieve this, we need to introduce some artificial delay since swiftUI doesn’t like it for some reason (and also a lot of times FaceID is so fast you don’t even notice the locking screen is there which is confusing and jarring )

Mlem/Extensions/View - BioLock.swift
```diff
@@ -0,0 +1,87 @@
...
+struct HandleAccountSecurity: ViewModifier {
+    @EnvironmentObject var accountsTracker: SavedAccountTracker
+    @EnvironmentObject var appState: AppState
+    let account: SavedAccount?
+    
+    @State var context = LAContext()
+    
+    func body(content: Content) -> some View {
+        if let account = account {
+            if accountsTracker.accountPreferences[account.id]?.requiresSecurity == true {
```
EricBAndrews, line 22: else condition is idempotent, condense these into a single check

JakeShirley:
Visually could we align this with the darkened blur + white text?

Mlem/Views/Tabs/Settings/Components/Views/General.swift
```diff
@@ -122,7 +141,55 @@ struct GeneralSettingsView: View
...
+                                if allowChangeLockState {
+                                    var oldSettings = accountsTracker.accountPreferences[account.id] ?? AccountPreference(requiresSecurity: false)
+                                    oldSettings.requiresSecurity = accountLocked
+                                    accountsTracker.accountPreferences.updateValue(oldSettings, forKey: account.id)
+                                    dirtyEditingUserAccount = false
+                                    return
+                                }
+                            } catch {}
+                            // if all went well we shouldn't get here
+                            //                               await MainActor.run {
```
JakeShirley, line 179: Looks like some left over debug code

JakeShirley (approved):
Minor nit but "Account Settings" makes it sound like it will only apply to a single account but these seem to apply cross-account.

EricBAndrews (changes requested):
One nit, everything else looks good

EricBAndrews:
> Visually could we align this with the darkened blur + white text?
> 

Agreed, that's a good idea

mormaer:
> This is not displayed using a full cover, I wanted the user to be able to go back, especially to the accounts, or the Settings menu if he’s ever in trouble, but be sure that the content itself has interactions disabled

I must say I disagree with this approach, any and all biometric lock I've seen implemented in iOS blocks all interaction with the application. FaceID/TouchID is a reliable system feature, and as long as we allow fallback to the users passcode they're not at risk of getting stuck unless they've lost their finger / face _and_ forgot their code 🙃 

What do other people think, should only part of the interface be obscured when locked?

I can see some value in allowing it to be set by account, but to me again we'd block the entire interface, if someone needed out they could always swap the app away and when it re-opens it will be on the account screen, tapping on a _protected_ account could prompt for biometrics before pushing into the user feed?

EricBAndrews:
We've been erring on the side of "do it the Apple way" pretty much every time it's come up, for consistency I'm going to agree with morm here

tht7:
I honestly don’t disagree. I think both approaches have a value to them, and I wanted to air on the side of caution (to me, that meant a lock that you can always escape from)

So I'll gladly change it

(I believe we should strive to be system, consistent and looking at a few examples I can confirm that the usual behavior is Face ID stop the interaction all together (so full cover), and when it comes to critical stuff when you can’t scan Face ID, or tap your passcode, it goes back to the previous interaction, which matches your description, so I’m also happy to change)

I will change the behavior for this feature in my next update

mormaer:
Screenshot 2023-06-28 at 23 10 20

here's a _disgusting_ draw.io fever-dream of how I would envision it working if we offer it per account 🙈 

I know, I know, I should be a designer 😂

ShadowJonathan:
Heya @tht7, we have a new repo at https://github.com/mlemgroup/mlem, due to some shuffling around we couldn't get it migrated properly, but if you could reopen this PR there, and link it back here, then it can continue to be considered ^^

Issue #308, opened 2023-06-26
Handle images in text posts


EricBAndrews:
Images in text posts should be scaled to fit and get a nice little border, just like image-only posts.

Child issues:

#291 
#262 
#266

EricBAndrews:
Closed in #336

Issue #291, opened 2023-06-25
Only the first NSFW image is censored


ShadowJonathan:
As in the title, also noted in #61

ShadowJonathan:
@EricBAndrews I can't yet affect the Development Tracker, but previously this was in Under Development

JakeShirley:
I just tried to reproduce this in a community feed but cannot.  All of the posts I am seeing are blurred.

Any tips?

ShadowJonathan:
TFF shows an "All" feed, not sure what else to try than that.

cc @BaelfireNightshd and @EricBAndrews since they might know more.

BaelfireNightshd:
I think my original test post was removed by moderation for some reason... I finally got another account and was going to try to get a screenshot, but since it might have been removed (I sorted by new and found a post in the mlemapp community that I made shortly before the test post and couldn't find the test post I made. I'll check on web when I get a chance to see if it's just the mlem client not showing it) it's hard to demonstrate short of finding an actual NSFW example (which I'm sure you can understand me being uncomfortable with sharing that on GitHub).

BaelfireNightshd:
I reached out to a mod to ask about the removal.

JakeShirley:
You could spin up a community and make an example post that is marked NSFW if you feel more comfortable that way.  All of the NSFW communities and posts I've seen seem to be working as far as I can tell.

BaelfireNightshd:
It's back up! https://beehaw.org/post/710425

Interestingly, none of the images are blurred. Maybe because I put text before? Maybe because the images are all in the body?

EricBAndrews:
Child issue of #308

BrooklynMan:
exactly. contents of the Description text box, including any images linked there, are NOT censored by the Lemmy API. We, however, should apply NSFW blur to all images there, too.

EricBAndrews:
Closed in #336

Issue #266, opened 2023-06-24
* Posts with multiple images cause resizing issues


MarsAgainstVenus:
**Describe the bug**
Scrolling through /c/all, a post with multiple images went from one image with a normal size to increasingly zoomed in images.

**Expected Behavior**
Each image should have been the same size as the first.

**Actual Behavior**
Images did not keep the correct size. 

**To Reproduce**
Scroll through @retrogaming@lemmy.world to this post: https://lemmy.world/post/518834 or any community with any post that contains more than one image.

**Mlem Information:**
0.1.0(14)

**Additional Context**
Any other relevant information

EricBAndrews:
Child issue of #308

EricBAndrews:
Closed in #336

Issue #262, opened 2023-06-24
* post goes partly off-screen


Emiliaaah:
**Describe the bug**
I came across this post on my feed https://lemmy.ml/post/1466711

**Expected Behavior**
I would expect the post to be entirely onscreen

**Actual Behavior**
A part of the post is not visible (you have to do some swipes magic to see part of the post).

https://github.com/mormaer/Mlem/assets/55017867/85a98e03-ebc8-4588-9c87-97f027703651

**To Reproduce**
You'd need this post to show up in your feed, or look for it in the piracy community (piracy@lemmy.dbzer0.com)

**Mlem Information:**
Version 0.1.0 (14)
I've only seen this happening on the new version but it could be that it's something unique to the post that caused this.

**Additional Context**
Any other relevant information

Emiliaaah:
I also found the following post https://lemmy.ml/post/1444424

https://github.com/mormaer/Mlem/assets/55017867/ac9e8b91-6978-4dfb-a2fd-64c9ce5fbfca

Emiliaaah:
And another one https://lemmy.dbzer0.com/post/35555

EricBAndrews:
Bundling this with #291--we need to handle images within text posts

EricBAndrews:
Child issue of #308

EricBAndrews:
Closed in #336

PR #336, opened 2023-06-27
Better-ish image handling

tht7:
# Checklist
      - *list issue(s) here*

# Pull Request Information

## About this Pull Request
This was a mess to handle but this should fix the NSFW blur and apply it on EVERY IMAGE in the app, when needed

created new CachedImageWithNsfwFilter that should be uniform in our app, it takes a url and uses our cache to load it, and blurs it according to the users wishes if you marked it as NSFW

introduced a new CachedImageProvider that can give you the uniform cached image view that's loaded from a URL and respects the nsfw setting 

Had to break the Markdown view since it wasn't flexible enough, so now the MarkdownView handles images before passing on the torch of actual markdown to the library's Markdown View

Related Issues:
#308 
#291 
#262 
#266 

## Screenshots and Videos
I won't 🤣


Mlem/Views/Shared/Markdown View.swift
```diff
@@ -236,9 +238,53 @@ struct MarkdownView: View
...
+            ForEach(0...max(blocks.count, images.count), id: \.hashValue) { i in
+                if blocks.count > i {
+                    getMarkdown(text: String(blocks[i]))
+                }
+                if images.count > i {
+                    imageProvider.makeImage(url: URL(string: String(images[i].1)))
+                }
+            }
+        }
+//        for i in 0...max(blocks.count, images.count) {
```
WestonHanners, line 277: is this commented code needed?

tht7: Nope! removed!

mormaer:
Hey @tht7 - this is nice 👍 

I don't have time to review tonight but will take a look tomorrow.

My first thoughts are that I'm wondering if we could clean it up slightly so that we always use the `CachedImageProvider`, given it can produce a view directly... it could give us a single way to treat images throughout the project if they went through a single entry point (and we could hide different behaviours behind it, not just NSFW filtering).

If we worked toward that then we could write a custom linting rule too to make sure everywhere uses it 👍

Mlem/Views/Shared/Markdown View.swift
```diff
@@ -201,8 +202,42 @@ private extension Color {
...
+            "!["
+            Capture {
+                ZeroOrMore(.any, .reluctant)
+            }
+            "]("
+            Capture {
+                ZeroOrMore(.any, .reluctant)
+            }
+            ")"
+        }
```
EricBAndrews, line 223: Nit, but for readability even simple regexes should have an explanatory comment

Mlem/Views/Shared/MarkdownImageProvider.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.url = url
+        self._showNsfwFilterToggle = .init(initialValue: true)
+    }
+
+    var body: some View {
+        ZStack {
+            CachedAsyncImage(url: url, urlCache: AppConstants.urlCache) { image in
+                image
+                    .resizable()
+//                    .frame(maxWidth: .infinity)
```
EricBAndrews, line 46: Can be removed?

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item Old.swift
EricBAndrews, line 1: This whole file should be removed, I think it snuck in with one of my PRs

EricBAndrews (changes requested):

WestonHanners (approved):

Mlem/Views/Shared/Markdown View.swift
tht7: Added comments!

Mlem/Views/Shared/MarkdownImageProvider.swift
tht7: removed!

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item Old.swift
tht7: ahaha it was there when I got it I swear! but removed as well 😄

tht7:
@mormaer actually, I think the opposite

 I created this image provider initially to comply with some protocol the markdown view laid out, but when I saw that the markdown view literally never uses that image provider (and trust me I desperately read the source code of that markdown view) I instantly dropped it and made a nice image _view_

I definitely think we should all unify and use the same image _view_ throughout the application. I definitely don't think this should be a provider.

Providers feel very anti-pattern in swift, they are less reactive, and anything they can do can be produced better and cleaner with simple views and the environment

So I completely removed that silly image provider

Mlem/Views/Shared/Cached Image With Nsfw Filter.swift
```diff
@@ -0,0 +1,77 @@
+//
+//  MarkdownImageProvider.swift
+//  Mlem
+//
+//  Created by tht7 on 26/06/2023.
+//
+
+import Foundation
+import SwiftUI
+import MarkdownUI
```
JakeShirley, line 10: Is this used here?

JakeShirley (approved):

Mlem/Views/Shared/Cached Image With Nsfw Filter.swift
EricBAndrews: Resolving for the sake of getting it merged, I'll check and fix if needed in upcoming post layout PR

tht7: Nooo it’s not! Dammit it’s a leftover

Issue #345, opened 2023-06-27
Generic Trackers


EricBAndrews:
Create a generic feed tracker and feed loading system to cut down on code duplication.

Core functionality:

- [x] Conforms to `ObservableObject`, publishing an array of items and an `isLoading` bool
- [x] Offers a `loadNextPage` method that loads the next page
- [x] Offers a `refresh` method that resets its items--it should fetch them into a temp variable before setting the main array to that in order to not display an empty feed while refreshing or if failed
- [x] Offers an `update` method that allows replacing an item with an updated version of that item (used for interactions)
- [x] Offers a way to easily identify whether a given item is the 10th-from-last in its array (used for loading infinite feed; note that this needs to be able to work in threaded feeds, so if an item is 10th-from-last in the *SomeItem* array but 50th-from-last in the (*SomeItem* interlaced with *AnotherItem*) array, it can still identify it as 10th-from-last and load accordingly--done with `loadMarkId` for inbox)

mormaer:
Implemented in #375

PR #375, opened 2023-06-29
345 - Generic trackers

mormaer:
# Checklist
      - #345

# Pull Request Information

## About this Pull Request
This PR addresses #345.

It introduces a generic `FeedTracker` class which can be typed to match items you expect to retrieve as a feed from the API such as Posts, Messages etc.

In order to use this tracker the API requests and types you wish to use must conform to the following protocols:

Request -> `FeedTrackerItemProviding`: This identifies an API call as being able to return a collection containing types which match the `FeedTrackerItem` protocol

Items -> `FeedTrackerItem`: This identifies a model as being a feed item and so can be understood and used generically within the feed trackers methods.

## Usage
Once you have updated a model and request to conform to the above protocols you can then create a new tracker with the following definition:

```swift
class ExampleTracker: FeedTracker<ExampleItem> {

func loadItems(account: ...) {
    try await perform(
        GetExampleItemRequest(account: ..., page: page)
    )
}
```

In the above example we have defined a single `loadItems` function which would take the contextual arguments required to fetch your items, such as the users account, sort types or any other relevant values for your use case. We also pass the `page` value which is defined by the tracker itself. It will always begin at `1` after initialisation.

The `perform(...)` function defined on the generic tracker takes your pre-built request and will do the heavy lifting of calling the API, adding the items (less duplicates) into it's items property and incrementing the `page` property ahead of your next call.

As the tracker manages it's own `page` please always use the defined value except in cases where you are looking to perform a refresh, in which case use the `refresh(...)` function which also takes a request, however you would explicitly set your page to `1`, the tracker will reset it's other internal values as appropriate since you have used the `refresh(...)` method.

In addition to this the tracker offers a `shouldLoadContent(after: ...)` method which will accept any of the items currently held within the tracker. This will return `true` when the tracker has determined it is appropriate to load the next page, this can be used in our views to load on request as the user scrolls the feed.

By default the tracker will attempt to perform a merge sort when retrieving new items, if you are building a tracker which is not returned in date order please pass an argument to the initialiser to disable this behaviour:
```swift

class ExampleTracker: FeedTracker<ExampleItem> { ... }

let tracker = ExampleTracker(shouldPerformMergeSorting: false)
```


Mlem/Feed/FeedTracker.swift
```diff
@@ -0,0 +1,121 @@
...
+    
+    /// A method to refresh this tracker, this will reset the state of the tracker to it's first page and set the retrieved items when returned
+    /// - Parameter request: An `APIRequest` that conforms to `FeedItemProviding` with an `Item` type that matches this trackers generic type
+    /// - Returns: The `Response` type of the request as a discardable result
+    @discardableResult func refresh<Request: APIRequest>(
+        _ request: Request
+    ) async throws -> Request.Response where Request.Response: FeedTrackerItemProviding, Request.Response.Item == Item {
+        let response = try await retrieveItems(with: request)
+        reset(with: response.items)
+        return response
```
EricBAndrews, line 69: Page should be incremented somewhere in here

EricBAndrews (approved):
Very clean

Issue #401, opened 2023-07-01
Scrolling replies/mentions/messages jumps to top


EricBAndrews:
Scrolling any of the detail views in inbox causes jump to the top when more items are loaded.

PR #403, opened 2023-07-01
Inbox Hotfix

EricBAndrews:
# Checklist
      - #401 

Applies an ugly little hotfix, issue exists for long-term resolution.

On-the-fly loading of inbox detail feeds (messages, replies, mentions) would trigger a re-render before the call came through, resulting in the scroll position being snapped to the top of the feed. This is likely caused by the shared trackers; long-term, careful state isolation should prevent this issue (probably instantiating the trackers at `InboxView`, then passing them down as `@Binding`s).

This hotfix sidesteps that by leveraging the functional operation of `allItemsFeed`; the added hotfix logic creates filtered sets of `InboxItems` and passes them to `allItemsFeed` depending on the selected feed.


Mlem/Views/Tabs/Inbox/Feed/All Items Feed View.swift
```diff
@@ -72,4 +72,28 @@ extension InboxView {
             }
         }
     }
+    
+    // TODO: no. just... no.
```
JakeShirley, line 76: 😁

JakeShirley (approved):

PR #405, opened 2023-07-01
Fix #32 - hide downvote button when disabled

### LINKED ISSUE #32: Hide downvote buttons for communities with feature disabled
## Introduction
Some instances of Lemmy have downvotes disabled community-wide. Mlem should hide the downvote buttons when viewing the content in these communities.

## Requirements
Recognize if downvotes are hidden in a specific Lemmy instance and hide buttons appropriately.

## Discussion
An example of a community with disabled downvotes: [Beehaw.org](https://beehaw.org)

mormaer:
# Checklist
      - #32

# Pull Request Information

## About this Pull Request
This PR addresses #32

As some instances choose to disable downvoting we should hide the option across posts and comments since the user cannot perform those actions. The changes in this PR detect when we're selecting an account and fires off a non-blocking call to retrieve the site information, from this we can determine if downvoting is disabled and update a new value in our `AppState`.

The vote complex views have been updated to read this value and when `false` the downvote button is not rendered.

## Screenshots and Videos
| BEFORE | AFTER |
| --- | --- |
| post_before | post_after |
| comment_before | comment_after |

## Additional Context

I had a whole other branch where I did this the other day, and came to write up the reasons I wasn't happy about the approach I'd come up with due to SwiftUI being awkward about letting me know when a user had tapped a row... while I was typing up my rant I ended up _rubber-ducking_ my own brain and realised there was a much cleaner/simpler way to approach it.

The changes from the improved version are included in this PR, but for the sake of transparency and also because I'm _still mad about it_ I've kept the rant below for your reading pleasure:

<details><summary>Rant</summary>There are a few things I'm not happy about with this approach 😓 

Firstly, performing an action in response to a selection appears harder than it should be in SwiftUI. I've altered our list rows on the account page so that they now respond to interaction via a `onTapGesture` which calls a separate function where we can perform additional actions (in this case a network call) as well as adding to our navigation stack.

The downside of this is that our account rows are no longer navigation links, so we lose the native disclosure icon on the trailing edge. I've added a _fake_ one back in, but that's horrible.

I tried a few other approaches such as deriving a `Binding<Bool>` from an optional property, and setting that property via a button as seen in various articles ([example](https://www.avanderlee.com/swiftui/navigationlink-programmatically-binding/)) however after implementing that... the disclosure was gone again 😧

I also realised that `.navigationDestination` wants a closure which returns a view, so I thought _"oh, I could trigger the call inside that closure as long as I then return the destination view!"_. That compiled, and worked... then I realised in my proxy that the call had been made about a _billion_ times and I'd gone over the rate limit on my instance within a split second. I reverted that.

If anyone knows how to respond to a user selection on a row that is a `NavigationLink` inside a `NavigationStack` please do tell me as I really want to know 😢</details>


Mlem/API/Models/Site/APILocalSite.swift
```diff
@@ -38,7 +38,7 @@ struct APILocalSite: Decodable {
 
 // lemmy_db_schema::source::local_site::RegistrationMode
 enum APIRegistrationMode: String, Codable {
-    case closed = "Closed"
-    case requireApplication = "RequireApplication"
-    case open = "Open"
+    case closed = "closed"
+    case requireApplication = "requireapplication"
+    case open = "open"
```
mormaer, line 43: Hey @ShadowJonathan - I had to make a couple of changes in this model as it was unable to decode, the first one `actorNameMaxLength` I can verify against the Rust side, but for this string based enum I was struggling to confirm why it was all lowercase in my responses (this was on `beehaw` and `test-mlem`) - are you able to comment on whether this is expected across all instances, or do we need to handle both `closed` and `Closed`?

ShadowJonathan: `#[serde(rename_all = "lowercase")]`

should be fine as-is, thanks for checking, and sorry for my tardiness

ShadowJonathan (approved):
LGTM, I see that this doesnt disable the gestures for downvoting though, would you want to handle that too?

mormaer:
Oh good catch @ShadowJonathan - I forgot about those entirely, I'll address those before we merge this 👍

mormaer:
> LGTM, I see that this doesnt disable the gestures for downvoting though, would you want to handle that too?

Hey @ShadowJonathan, I've popped up a [commit](https://github.com/mormaer/Mlem/pull/405/commits/c6367ce3ae2476ead0aac5c8601b6d9096f9a509) that removes the long left action (downvote) in this scenario... it's a bit rough and ready but to do it nicer involves a bigger refactor to the Swipe Actions.

@EricBAndrews - if you don't mind I may take a look at refactoring them a bit to allow for optionality in a _nicer_ way than in my above commit?

EricBAndrews:
Go for it--they're due for a refactor anyhow, and having a little more customization also pushes the ball forward on inbox interactions and 1.2

mormaer:
> Go for it--they're due for a refactor anyhow, and having a little more customization also pushes the ball forward on inbox interactions and 1.2. While you're at it, it would be *lovely* if they didn't require a stack of parameters as half mile thick

Will do, aye, I was going to bundle each action up into a lil' struct or such, it should help reduce the parameter count, could perhaps define static vars for them after that's done so you can pass the action like `.upvote(callback)` 🤔

That _should_ like you said make it easier to customise which actions are ok which sides for each user 🙏

Mlem/App State.swift
```diff
@@ -27,4 +26,21 @@ class AppState: ObservableObject {
...
+    
+    func setActiveAccount(_ account: SavedAccount?) {
+        currentActiveAccount = account
+        
+        if let newAccount = account {
+            Task {
+                let request = GetSiteRequest(account: newAccount)
+                if let response = try? await APIClient().perform(request: request) {
+                    await MainActor.run {
+                        enableDownvote = response.siteView.localSite.enableDownvotes
```
JakeShirley, line 40: After talking about how we will need language data I wonder if we can just cache the entire site view.

JakeShirley (approved):
We should make sure to include a blerb in the changelogs about why the downvote arrow may "disappear" for some users.

Mlem/App State.swift
mormaer: We should definitely store them here once we know how we're going to use them, to save doing a separate call 👍

PR #415, opened 2023-07-02
Refactor - swipe actions

mormaer:
# Checklist
      - Not raised as a specific issue, but as discussed [here](https://github.com/mormaer/Mlem/pull/405#issuecomment-1615939874)

# Pull Request Information

## About this Pull Request
This PR refactors our swipe actions to improve readability and allow for better optionality handling for actions.

- Introduced a `SwipeAction` struct which houses the values that drive each action
- Refactored the code to use this new struct which reduces the number of parameters which must be passed directly to the modifiers initialiser
- Improved support for optionality in the gestures
    - if you do not pass a `secondary` action then only the `primary` will display
    - if you do not pass a `primary` action then a swipe from that edge will be ignored
    - a small `assert` has been added to flag during development if we're passing a `secondary` without a `primary`
- Moved creation of the comment/post actions into extensions to keep the main body of the views less cluttered

## Additional Context
This improves the API of adding swipe actions to our views, as we can now easily support single actions from the leading/trailing edges. We can also choose to supply no actions on one edge, but still support them on the other.

I've left a `TODO` for aligning the behaviour of the icons between comments and posts, comments currently support showing a _undo_ style icon when the action will reset your vote/save etc, but due to the structure of the code in posts this is not possible without a larger change.

This would be best approached after we tackle #117 and properly unify the voting actions.

### DIFF
--- a/Mlem/Extensions/Swipey Actions.swift
+++ b/Mlem/Extensions/Swipey Actions.swift
@@ -46,6 +46,16 @@ struct SwipeyView: ViewModifier {
          primaryTrailingAction: SwipeAction?,
          secondaryTrailingAction: SwipeAction?
     ) {
+        assert(
+            secondaryLeadingAction != nil && primaryLeadingAction == nil,
+            "No secondary action should be present without a primary"
+        )
+        
+        assert(
+            secondaryTrailingAction != nil && primaryTrailingAction == nil,
+            "No secondary action should be present without a primary"
+        )
+        
         self.primaryLeadingAction = primaryLeadingAction
         self.secondaryLeadingAction = secondaryLeadingAction
         self.primaryTrailingAction = primaryTrailingAction
--- a/Mlem/Extensions/Swipey Actions.swift
+++ b/Mlem/Extensions/Swipey Actions.swift
@@ -8,7 +8,13 @@
 import SwiftUI
 
 struct SwipeAction {
-    let symbolName: String
+    
+    struct Symbol {
+        let emptyName: String
+        let fillName: String
+    }
+    
+    let symbol: Symbol
     let colour: Color
     let action: () async -> Void
 }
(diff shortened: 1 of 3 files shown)


WestonHanners (approved):
image

Mlem/Extensions/Swipey Actions.swift
```diff
@@ -7,84 +7,63 @@
...
 
+struct SwipeAction {
+    
+    struct Symbol {
+        let emptyName: String
+        let fillName: String
+    }
+    
+    let symbol: Symbol
+    let colour: Color
```
EricBAndrews, line 18: Nit but we should probably go American with the spelling because that's how the library spells it

Mlem/Extensions/Swipey Actions.swift
```diff
@@ -126,68 +105,45 @@ struct SwipeyView: ViewModifier {
...
-                            }
-                        }
-                        // bounce back to neutral
-                        withAnimation(.interactiveSpring()) {
-                            dragPosition = .zero
-                            leftSwipeSymbol = emptyLeftSymbolName
-                            rightSwipeSymbol = emptyRightSymbolName
-                            dragBackground = .systemBackground
-                        }
+                        draggingDidEnd()
```
EricBAndrews, line 108: Nice

Mlem/Extensions/Swipey Actions.swift
```diff
@@ -202,63 +158,72 @@ struct SwipeyView: ViewModifier {
...
+        if position > 0, primaryLeadingAction == nil {
+            return false
+        }
+        
+        if position < 0, primaryTrailingAction == nil {
+            return false
+        }
+        
+        return true
+    }
```
EricBAndrews, line 206: Really like this function, it's going to make it easier to fine-tune drag detection too

EricBAndrews (changes requested):

Mlem/Extensions/Swipey Actions.swift
mormaer: ah yeah, muscle memory - I'll sort it 👍

mormaer: also, _American_ is a funny way to spell _wrong_ 😉

mormaer: sorted, added an `unfix` commit 😉

EricBAndrews:
Dupe #420

